PR #13, opened 2017-04-30
fix travis shield

yuce:


yuce:
Merging this myself, since it's trivial.

PR #19, opened 2017-05-01
Cache options

yuce:


orm.go
```diff
@@ -453,3 +469,19 @@ const (
 	TimeQuantumMonthDayHour     TimeQuantum = "MDH"
 	TimeQuantumYearMonthDayHour TimeQuantum = "YMDH"
 )
+
+type CacheType string
```
jaffee: What's the point of declaring this as separate type?

yuce: Isn't that the idiomatic way of having enums? http://golang-basic.blogspot.com.tr/2014/07/step-by-step-guide-to-declaring-enums.html

jaffee: That makes sense if the value is an int, and/or you want some type safety around it or to define methods on it, but I don't think it buys us anything here - users are going to use the constant values when declaring cache types - or they'll just be annoyed that they have to wrap a string in `pilosa.CacheType()` - so you'll have a very redundant looking line like:

```
CacheType: pilosa.CacheType("ranked"),
```

yuce: What's wrong with `CacheType: CacheTypeRanked` ? This is actually in the tests:
```

Comment (jaffee) [inline orm.go]: yea, I'm totally fine with that - but there's no reason for it to have a type other than `string` in that case.

I was just trying to make the point that defining the `CacheType` type wasn't adding anything useful.

Comment (yuce) [inline orm.go]: I think that's useful as documentation at the least (didn't check whether there's some additional special IDE/tooling support for that). `FrameOptions` is defined like:
```
type FrameOptions struct {
	RowLabel string
	TimeQuantum TimeQuantum
	InverseEnabled bool
	CacheType      CacheType
	CacheSize      uint
}
```
So that immediately tells that `CacheType` field is of type `CacheType` (not just a generic string or int).

Also that serves as lightweight data encapsulation as well. If we one day change the underlying data type from string to int (long shot), the user code don't need to break.

Comment (jaffee) [inline orm.go]: The data encapsulation argument doesn't hold water - user code would look like `CacheType("blah")` which would just be a compile error if CacheType was changed to int. If the user code was using the constants, then it would continue working fine either way.

If we really wanted to encapsulate, we could make `cacheType` an unexported struct type with 1 string field, and only export variables of that type, so that user code is forced to use the variables since it cannot construct arbitrary values of that type. I think that'd be overkill though.

In any case, if you want to keep it as a separate type, I disagree, but I'll support it.

File: orm.go [hunk 2]
```diff
@@ -453,3 +469,19 @@ const (
...
+// CacheType constants
+const (
+	CacheTypeDefault CacheType = ""
+	CacheTypeLRU     CacheType = "lru"
+	CacheTypeRanked  CacheType = "ranked"
+)
+
+func encodeMap(m map[string]interface{}) string {
+	result, _ := json.Marshal(m)
+	// ignoring the error, since this function is package private
```
jaffee: If this is expected never to generate a non-nil error, we should panic if it does rather than silently ignoring it.

orm.go
yuce: I don't understand why the user would type `CacheType("lru")` rather than `CacheTypeLRU`.

jaffee: I hope they wouldn't! If they did though, then it would fail if we changed the type of CacheType.

If they use the constant as you suggest, then we can change the type of the constant regardless of whether it is aliased to `CacheType` or a bare int or string or anything else.

orm.go
yuce: Updated

orm.go
yuce: No, suppose the user defined a function which takes an arg of `CacheType` or a struct with `CacheType` type:
```
type MyStruct struct {
CacheType pilosa.CacheType
}
```
Even if we change `CacheType` from `string` to `int`, the code wouldn't break (if they use the `CacheType("lru")` style, they are on their own). If `CacheType` was defined as a string, obviously their code would break.

yuce: Anyway, IMO the documentation aspect is more important, since `Something CacheType` conveys more information than `Something string`.

PR #26, opened 2017-05-24
sync validators with the client spec and server

yuce:


yuce:
Client spec is at: https://github.com/yuce/general/blob/client-spec/proposal/client-design.md

PR #27, opened 2017-05-25
Rename db->index in tests and readme

alanbernstein:
This is every occurrence of "db" except for those in protobuf files - I'm not sure if those can/should be changed here?

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ var err error
 client := pilosa.DefaultClient()
 
 // Create an Index object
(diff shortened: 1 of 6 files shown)


travisturner:
@yuce @alanbernstein we should update the `public.proto` definition file to be consistent with the server (change "DB" to "Index" in `ImportRequest`) and regenerate the `public.pb.go` file.

yuce:
@travisturner If it's OK I will defer updating the proto file since Go client doesn't support importing yet (and the proto file may change until we implement that)

PR #28, opened 2017-05-25
Require minimum args to bitmap operations through explicit check inst…

alanbernstein:
…ead of through function signature


yuce:
This looks good!

PR #29, opened 2017-06-09
Updates for v0.4.0. Added more tests

yuce:


orm.go
```diff
@@ -391,28 +388,57 @@ func (f *Frame) ClearBit(rowID uint64, columnID uint64) *PQLBaseQuery {
...
+// InverseTopN creates a TopN query with the given item count.
+// Returns the id and count of the top n bitmaps (by count of bits) in the frame.
+// This variant sets inverse=true
+func (f *Frame) InverseTopN(n uint64) *PQLBitmapQuery {
+	return NewPQLBitmapQuery(fmt.Sprintf("TopN(frame='%s', n=%d, inverse=true)", f.name, n), f.index, nil)
 }
 
 // BitmapTopN creates a TopN query with the given item count and bitmap.
-// This variant supports customizing the bitmap query.
+// This variant supports customizing the bitmap query and sets inverse=true.
```
travisturner: shouldn't this be `inverse=false`?

client_it_test.go
```diff
@@ -179,6 +180,67 @@ func TestSetRowAttrs(t *testing.T) {
...
+	}
+	qry := index.BatchQuery(
+		frame.SetBit(10, 5),
+		frame.SetBit(10, 10),
+		frame.SetBit(10, 15),
+		frame.SetBit(20, 5),
+		frame.SetBit(30, 5),
+	)
+	client.Query(qry, nil)
+	time.Sleep(10 * time.Second)
```
travisturner: It might be a good idea to include a comment describing why this sleep is here. We might also consider exposing a way for the client to request a cache invalidation, so in that case the test could call that instead of sleeping.

README.md
```diff
@@ -173,13 +176,18 @@ Index:
 Frame:
 
 * `Bitmap(rowID uint64) *PQLBitmapQuery`
+* `InverseBitmap(rowID uint64) *PQLBitmapQuery`
```
travisturner: should `rowID` be `columnID` instead?

client_it_test.go
yuce: I've added a comment for other clients but missed this one. Thanks for noticing. I think it would be great to be able to cause the Pilosa server to invalidate caches/flush the data. Apart from testing, that would be very desirable when/if we incorporate 3rd party backup solutions. I was thinking maybe we could send a special signal to the server which causes invalidation/flush (but that isn't easily usable by tests).

README.md
yuce: Fixing

orm.go
yuce: Fixing

yuce:
@travisturner Updated with the fixes.

PR #31, opened 2017-06-19
Added support for imports

yuce:
The coverage dropped to 99%, I will add a few more tests.


imports.go
```diff
@@ -0,0 +1,135 @@
...
+	return &CSVBitIterator{
+		reader: reader,
+		line:   0,
+	}
+}
+
+// Iterate iterates on all lines of a Reader.
+// If the callback returns false, the iteration is paused; call it again to resume.
+// Returns io.EOF on end of iteration
+func (c *CSVBitIterator) Iterate(callback BitIteratorCallback) error {
```
jaffee: instead of passing in a callback, why not just have this be like a `Next() (Bit, error)` function - this is how most iterators are done in Go in my experience.

yuce: Java has a well defined interface for iterators, but I didn't find something similar in the Go stdlib. I was trying to decouple the iterator from the code executed on iteration but I am also not very happy with the result. I'll have another look to this bit of code.

yuce:
@jaffee replaced the iterator callback with the next pattern

yuce:
@travisturner @jaffee does this PR look OK to merge ?

client.go
```diff
@@ -181,6 +183,94 @@ func (c *Client) Schema() (*Schema, error) {
...
+func (c *Client) ImportFrame(frame *Frame, bitIterator BitIterator, batchSize uint) error {
+	const sliceWidth = 1048576
+	linesLeft := true
+	bitGroup := map[uint64][]Bit{}
+	var currentBatchSize uint
+	indexName := frame.index.name
+	frameName := frame.name
+
+	for linesLeft {
+		bit, err := bitIterator.NextBit()
```
jaffee: I would name the method `Next()` - `bitIterator.NextBit()` is somewhat redundant.

yuce: I got the inspiration from: https://godoc.org/google.golang.org/api/iterator . They named the iterator `NextPage`.

client_it_test.go
```diff
@@ -502,6 +537,153 @@ func TestInvalidSchema(t *testing.T) {
 	}
 }
 
+func TestCSVImport(t *testing.T) {
+	client := getClient()
+	text := `10,7
```
jaffee: is there a test like this one, but with timestamps?

yuce: There's a unit test with timestamps in `imports_test.go`.

jaffee: woops, I was adding single comments instead of reviewing - just found that.

PR #33, opened 2017-06-28
Status and Schema

yuce:
I just need to add the `client.Sync` function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,11 @@ Go client for Pilosa high performance distributed bitmap index.
 
 ## Change Log
 
+* **next**:
+    * Supports imports.
+    * Introducing `client.Schema` which retrieves the schema information from the server side. No need to re-define already existing indexes and frames.
+    * **Deprecation** `NewIndex`. Use `schema.Index` instead.
+
 * **v0.4.0** (2017-06-09):
     * Supports Pilosa Server v0.4.0.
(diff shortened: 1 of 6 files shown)


yuce:
I've added `client.SyncSchema` which deprecates `Create*` and `Ensure*` functions

README.md
```diff
@@ -52,17 +58,17 @@ var err error
...
 
 // Create a Frame object
 myframe, err := myindex.Frame("myframe", nil)
 
-// Make sure the frame exists on the server
-err = client.EnsureFrame(myframe)
+// make sure the index and frame exists on the server
+err = client.SyncSchema(schema)
 
 // Send a SetBit query. PilosaException is thrown if execution of the query fails.
```
jaffee: Minor comment - rather than "PilosaException is thrown", we should really be saying "PilosaError is returned" since Go doesn't have the notion of throwing exceptions.

yuce: I've copied quite a bit between the clients especially the documentation. I'll fix this.

yuce:
The decoupling between the ORM and the client helps a lot, we write unit tests instead of integration tests; we only have logging in the client code, etc. Even if we enabled creating the index in `schema.Index` call, we would have to link a client to the `Schema` object or to the `Index` call, which would remove the coupling. Also that would cause a create index call to the server every time (which would be unwanted especially if the index already exists on the server side).

PR #34, opened 2017-06-29
convenience method for creating new cluster from string slice

jaffee:
Creating a client against a cluster is a bit verbose without this.


yuce:
An update to the *next* section in the change log in the README would be great.

jaffee:
good call @yuce That's updated

yuce:
Great, thanks!

PR #35, opened 2017-07-03
make query response marshal to reasonable json

jaffee:
Came across this while working on benchmarking tools. Want to include the output as json, but results and columns weren't included since they weren't exported fields.

I could have written a custom MarshalJSON function instead of change the names of the fields, but it would have been a lot more work, and probably would have resulted in unnecessary allocations, so I went with what seemed like the simplest approach.

### DIFF
--- a/response.go
+++ b/response.go
@@ -40,10 +40,10 @@ import (
 
 // QueryResponse represents the response from a Pilosa query.
 type QueryResponse struct {
-	results      []*QueryResult
-	columns      []*ColumnItem
-	ErrorMessage string
-	Success      bool
+	ResultList   []*QueryResult `json:"results,omitempty"`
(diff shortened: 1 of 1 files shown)


yuce:
`ResultList` and `ColumnList` make `Result()`, `Results()` and `Columns()` obsolete. We can deprecate  and remove them in a later release. My only worry is, whether that would paint us in a corner in case we later want to create result objects on demand.

jaffee:
I agree it's nice to have them be behind functions for normal usage - I'd say let's not deprecate them, in fact, I think I'd be in favor of hiding QueryResponse behind an interface with those methods. I think json marshalling will still work, and we can more likely guarantee compatibility going forward even as we add things like streaming results. How's that sound?

PR #37, opened 2017-07-14
Added export support

yuce:


client.go
```diff
@@ -498,6 +529,24 @@ func bitsToImportRequest(indexName string, frameName string, slice uint64, bits
 	}
 }
 
+func statusToNodeSlicesForIndex(status *Status, indexName string) map[uint64]*URI {
```
travisturner: it's always a red flag for me when i see something nested more than one (or maybe two) levels. i understand why you've nested this series of for loops so deeply, but it still bothers me. at the very least i would change `if index.Name == indexName {...}` to `if index.Name != indexName { continue }` to remove one level of nesting.

also, are you ok ignoring errors from `NewURIFromAddress`?

finally, even though this is an unexported function I think it would be a good idea to add a comment to explain what it's for.

client.go
```diff
@@ -578,3 +627,53 @@ type StatusMeta struct {
...
+func newExportReader(sliceURIs map[uint64]*URI, frame *Frame, view string) *exportReader {
+	return &exportReader{
+		sliceURIs:  sliceURIs,
+		frame:      frame,
+		sliceCount: uint64(len(sliceURIs)),
+		view:       view,
+	}
+}
+
+func (r *exportReader) Read(p []byte) (n int, err error) {
```
travisturner: let's be sure to include comments for all exported functions.

client.go
```diff
@@ -578,3 +627,53 @@ type StatusMeta struct {
...
+		}
+		r.bodyIndex = 0
+	}
+	n = copy(p, r.body[r.bodyIndex:])
+	r.bodyIndex += n
+	if n >= len(r.body) {
+		r.body = nil
+		r.currentSlice++
+	}
+	return
```
travisturner: this function seems to have a mixture of named/unnamed return values. first, we should stick with one or the other within a function. second, when using named return values the intention is to make it clear what the return values means. in this case, the name `n` doesn't really tell me much about that value so it's not very useful. you might as well have them unnamed.

client.go
```diff
@@ -344,31 +349,56 @@ func (c *Client) fetchFragmentNodes(indexName string, slice uint64) ([]FragmentN
...
-	_, _, err := c.httpRequest("POST", "/import", data, noResponse)
+	_, _, err := c.httpRequest("POST", "/import", data, protobufHeaders, noResponse)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
+// ExportFrame exports bits for a frame
```
travisturner: comments should end with a period.

client.go
yuce: `NewURIFromAddress` would not fail in this case, as long as the the valid URI format is the same between the server and the client but I can return an error here too (omitting it saved us an extra test).

client.go
yuce: The `Read` function's signature is direct from Go docs: https://golang.org/pkg/io/#Reader so I thought its value would be obvious. Removed unnamed return values.

yuce:
@travisturner Updated.

client.go
travisturner: i don't mind so much that you're swallowing the error as long as that was your intent. perhaps you could leave a comment that you intentionally ignore the error because they are assumed to be the same.

client.go
```diff
@@ -665,7 +674,7 @@ func (r *exportReader) Read(p []byte) (n int, err error) {
 			r.frame.index.Name(), r.frame.Name(), r.currentSlice, r.view)
 		_, r.body, err = client.httpRequest("GET", path, nil, headers, errorCheckedResponse)
 		if err != nil {
-			return 0, err
+			return
```
travisturner: i guess my comment was more that the naked return can be confusing, especially in a longer function: https://tour.golang.org/basics/7

client.go
```diff
@@ -539,6 +539,7 @@ func statusToNodeSlicesForIndex(status *Status, indexName string) map[uint64]*UR
 			}
 			for _, slice := range index.Slices {
 				uri, err := NewURIFromAddress(node.Host)
+				// err will always be nil, but prevent a panic in the odd chance the server returns an invalid URO
```
travisturner: is `URO` correct?

client_it_test.go
```diff
@@ -982,6 +982,42 @@ func TestFetchViewsUnmarshalFails(t *testing.T) {
 	}
 }
 
+func TestStatusToNodeSlicesForIndex(t *testing.T) {
+	// even though this function isn't reallyu an integration test,
```
travisturner: typo with "reallyu"

yuce:
Thanks, fixed the typos.

PR #40, opened 2017-07-31
Updated the readme with import/export info

yuce:


README.md
```diff
@@ -356,6 +356,96 @@ countItems := response.CountItems
 count := response.Count
 ```
 
+## Importing and Exporting Data
+
+### Importing Data
+
+If you have large amounts of data, it is more efficient to import it to Pilosa instead of several `SetBit` queries. This library supports importing bits of a frame. The target frame must have been created before hand.
```
travisturner: If you have large amounts of data, it is more efficient to import it into Pilosa instead of using multiple `SetBit` queries. This library supports importing bits into an existing frame.

README.md
```diff
@@ -356,6 +356,96 @@ countItems := response.CountItems
...
+
+Before starting the import, create an instance of a struct which implements `BitIterator` and pass it to `client.ImportFrame` function. This library ships with `CSVBitIterator` struct which supports importing bits in the CSV (comma separated values) format:
```
travisturner: Before starting the import, create an instance of a struct which implements `BitIterator` and pass it to the `client.ImportFrame` function. This library ships with the `CSVBitIterator` struct which supports importing bits in CSV (comma separated values) format:

README.md
```diff
@@ -356,6 +356,96 @@ countItems := response.CountItems
...
+```
+ROW_ID,COLUMN_ID
+```
+
+Optionally, a timestamp with GMT time zone can be added:
```
travisturner: should this be `UTC` instead of `GMT`?

travisturner: actually, I'm not sure we should be specifying timezone at all; Pilosa is not aware of time zone.

README.md
```diff
@@ -356,6 +356,96 @@ countItems := response.CountItems
...
+```
+ROW_ID,COLUMN_ID,TIMESTAMP
+```
+
+Note that, each line corresponds to a single bit and the lines end with a new line (`\n` or `\r\n`).
```
travisturner: Note that each line corresponds to a single bit and ends with a new line (`\n` or `\r\n`).

README.md
```diff
@@ -356,6 +356,96 @@ countItems := response.CountItems
...
+    if it.index < len(it.bits) {
+        return Bit{}, io.EOF
+    }
+    return it.bits[it.index++], nil
+}
+```
+
+### Exporting Data
+
+You can export a view of a frame from Pilosa using `client.ExportFrame` function which returns a `BitIterator`. use `NextBit` function of this iterator to receive all bits for the specified frame. When there are no more bits, `io.EOF` is returned.
```
travisturner: You can export a view of a frame from Pilosa using the `client.ExportFrame` function which returns a `BitIterator`. Use the `NextBit` function of this iterator to receive all bits for the specified frame. When there are no more bits, `io.EOF` is returned.

README.md
```diff
@@ -356,6 +356,96 @@ countItems := response.CountItems
...
+
+The `PilosaClient` struct has the `Views` functions which returns views of frame. You can use that function to retrieve view names:
```
travisturner: The `PilosaClient` struct has the `Views` function which returns all of the views for a particular frame. You can use this function to retrieve view names:

README.md
yuce: UTC is equivalent to GMT, but it seems UTC is the newer name, so I will use it instead. Most of the time the programmer has to select between getting the timestamp in local time or UTC, so I think we should have a convention.

yuce:
Updated.

PR #44, opened 2017-08-06
fix link to Contributor License Agreement in the README

### LINKED ISSUE #43: Link to developer agreement is broken
The link in the README to https://wwww.pilosa.com/developer-agreement is broken.

travisturner:
Fixes #43

### DIFF
--- a/README.md
+++ b/README.md
@@ -451,10 +451,9 @@ for {
 
 Please check our [Contributor's Guidelines](https://github.com/pilosa/pilosa/CONTRIBUTING.md).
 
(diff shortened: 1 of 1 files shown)


travisturner:
ah, good point. thanks for catching that @codysoyland 
should we just remove that CLA bullet altogether?

codysoyland:
Yeah, I would be for removing it for now. The CLA currently seems to be specific to the Apache license, and these client libraries are under a different license, so we should discuss what to do.

PR #45, opened 2017-08-07
fixed issue with requiring an (optional) bitmap for TopN call

bmuller:
As well as adding the field name `filters` to the `TopN` call when arguments are given in the `FilterFieldTopN` call, per [the query language docs](https://www.pilosa.com/docs/query-language/).  Otherwise this error occurs:

```
Error: expected argument key, found "[" occurred at line 1, char 62
```

Updated tests for both fixes.


codysoyland:
Thanks again @bmuller! Looks good!

PR #46, opened 2017-08-11
WIP: add re-use for client connection on import  

tgruben:
added "caching" for direct client connections in order to more efficiently handle keep-alive connections

### DIFF
--- a/client.go
+++ b/client.go
@@ -314,8 +314,8 @@ func (c *Client) ImportFrame(frame *Frame, bitIterator BitIterator, batchSize ui
 	return nil
 }
 
-func (c *Client) getDirectClient(indexName string, slice uint64, host string) (*Client, error) {
-	key := fmt.Sprintf("%s:%d:%s", indexName, slice, host)
+func (c *Client) getDirectClient(indexName string, host string) (*Client, error) {
+	key := fmt.Sprintf("%s:%s", indexName, host)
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -61,6 +61,8 @@ type Client struct {
 	cluster *Cluster
 	host    *URI
 	client  *http.Client
+	slice_clients map[string]*Client
```
travisturner: can we call this something like `clientCache` (also, underscores are not typically used in go)

and with that change, in `getDirectClient()` I would change `cacheClient` to just `client` and change `newClient :=` to `client =`.

jaffee:
I wrote up a separate issue for some (potentially) more invasive changes, but really there should only be one net/http.Client object per host in the cluster, and it should be re-used for all requests to that host throughout the life of the program: https://github.com/pilosa/go-pilosa/issues/48

yuce:
The test for the cache is here: https://github.com/tgruben/go-pilosa/pull/1

yuce:
I think we should take the client cache out of the client and make it its own struct, something like `ClientManager`. Otherwise the address of client object which contains the cache is not in the cache:
```go
client := DefaultClient()
client2, _ := client.getDirectClient("foo.bar:10101")
if client != client2 {
    // client itself is not cached
}
```

PR #50, opened 2017-08-21
convert from glide to dep for dependency management

jaffee:


yuce:
The CI job fails. Should `.travis.yml` adapted for dep?

jaffee:
good call - I fiddled with that a bit and looks like it worked @yuce

jaffee:
Those are the generated examples - I think it's ok to leave them

PR #51, opened 2017-08-22
handle all errors, use pkg/errors to add context

### LINKED ISSUE #49: Client.ImportNode ignores error
```
 func (c *Client) importNode(request *internal.ImportRequest) error {
	data, _ := proto.Marshal(request)
	// request.Marshal never returns an error
```

`proto.Marshal` attempts to cast `request` to a `Marshaler`, but `*internal.ImportRequest` doesn't implement `Marshal()`,  so `proto.Marshal` goes down another path which it appears may return an error. We should probably just handle the error anywhere `proto.Marshal` is called regardless of whether the object in question is a `Marshaler` in order to be consistent, suppress warnings, and protect against future changes to generated code.

jaffee:
Fixes #49

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -7,9 +7,15 @@
   packages = ["proto"]
   revision = "888eb0692c857ec880338addf316bd662d5e630e"
 
+[[projects]]
+  name = "github.com/pkg/errors"
+  packages = ["."]
+  revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
+  version = "v0.8.0"
+
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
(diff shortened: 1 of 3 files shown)


yuce:
It would be great to have tests to keep the coverage at 100%. I can add them too.

jaffee:
hey @yuce - I can totally add the tests, I've got at least one more PR coming that may make some more invasive changes - do you mind if I wait until then to get the coverage back up to 100?

yuce:
@jaffee  Sure! I guess the incoming changes build upon this PR? Would it make sense to include this PR on the coming PR ?

yuce:
@jaffee I think https://github.com/pilosa/go-pilosa/pull/61covers everything in this PR. Is it OK to close this ?

jaffee:
yep, good call

PR #57, opened 2017-09-14
WIP: adds support for importing values via the `import-value` endpoint.

travisturner:
Added a new import function: `client.ImportValueFrame`.
The following code imports data into field `foo` in index `i`, frame `f`.

```
    client := pilosa.DefaultClient()

    // Retrieve the schema
    schema, err := client.Schema()

    text := `10,7
        11,5
        2,3
        7,1`
    iterator := pilosa.NewCSVValueIterator(strings.NewReader(text))

    index, _ := schema.Index("i", nil)
    frame, _ := index.Frame("f", nil)
    field := "foo"

    err = client.ImportValueFrame(frame, field, iterator, 10000)
    if err != nil {
        panic(err)
    }
```

TODO:
- [ ] Write tests.
- [ ] Update the documentation.


travisturner:
@yuce One thing I noticed is that my generated protobuf file seems to be different than what was there. I wonder if we're not using the same version of protoc? We should do something similar to the main pilosa repo where `make generate` specifies a version.

jaffee:
is there already a way to set up BSI fields through the go-pilosa client?

yuce:
@travisturner My protoc seems to be version 3.1.0:
```
$ protoc --version
libprotoc 3.1.0
```
As far as I see the version is not specified in the Pilosa Makefile too, but I guess there shouldn't be any problems as long as the client builds and tests suceed.

yuce:
@jaffee the client supports adding BSI fields like:
```go
frameOptions := &FrameOptions{
		RowLabel:       "stargazer_id",
		TimeQuantum:    TimeQuantumDayHour,
		InverseEnabled: true,
		CacheType:      CacheTypeRanked,
		CacheSize:      1000,
	}
err := frameOptions.AddIntField("foo", 10, 100)
```

jaffee:
perfect, thanks @yuce

PR #59, opened 2017-09-20
Response defaults + Custom time format

yuce:
The tests run fine locally against a custom built pilosa server with `field-value-import` branch. The CI build will fail.


yuce:
Thanks

PR #60, opened 2017-09-20
error handling and single http client

jaffee:
this combines a bunch of error handling, and making pilosa.Client only use a single http.Client object. It greatly simplifies pilosa.Client and makes cluster.Host the chief determiner of which host in the cluster should be queried (for queries where a specific host is not specified).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 all: test
 
 cover:
-	go test -cover -tags=integration -coverprofile=coverage.out
+	go test -cover -tags=integration
 
 generate:
 	protoc --go_out=. internal/public.proto
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 all: test
 
 cover:
-	go test -cover -tags=integration
+	go test -cover -tags=integration -coverprofile=coverage.out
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -675,10 +690,10 @@ func (options *ClientOptions) withDefaults() (updated *ClientOptions) {
 	*updated = *options
 	// impose defaults
 	if updated.SocketTimeout <= 0 {
-		updated.SocketTimeout = time.Second * 300
+		updated.SocketTimeout = time.Second * 10000
```
yuce: Isn't this timeout limit too high?

Makefile
```diff
@@ -4,7 +4,7 @@
 all: test
 
 cover:
-	go test -cover -tags=integration
+	go test -cover -tags=integration -coverprofile=coverage.out
```
yuce: Can we remove this? I'll update the Makefile and `make.cmd` so those flags will be changeable by the user.

client.go
jaffee: oops, sorry, that snuck in from some changes I made to support another project

PR #68, opened 2017-10-02
rename internal package to avoid name conflicts with protobufs

jaffee:
Tools like `pi` in `pilosa/tools` import both go-pilosa and pilosa. This causes some protobuf warnings like:
```
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.Bitmap
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.Pair
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.Bit
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.ColumnAttrSet
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.Attr
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.AttrMap
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.QueryRequest
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.QueryResponse
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.QueryResult
2017/10/02 14:47:06 proto: duplicate proto type registered: internal.ImportRequest
```

For now, these are mostly harmless, but probably confusing for new users.  The problem is that protobufs have a global namespace for packages, and so registering two types with the same package/name is an error. 

I'm open to suggestions for better naming. I think the other alternative would be to import the protobufs definitions from a shared package - I'm not opposed to this, but I figured I could make this  change quickly without involving other repos.


yuce:
@jaffee Updating last line of `.travis.yml` so it is `-ignore "pbuf/public.pb.go"` instead of `-ignore "internal/public.pb.go"` should fix the decreased coverage.

PR #74, opened 2017-10-31
Add QueryOptions.Slices support.

benbjohnson:
This commit adds support for specifying individual slices when executing a query:

	client.Query(bitmap, &QueryOptions{Slices: []uint64{0, 3}})

Fixes https://github.com/pilosa/pilosa/issues/641

### DIFF
--- a/client.go
+++ b/client.go
@@ -657,6 +657,7 @@ func newHTTPClient(options *ClientOptions) *http.Client {
 func makeRequestData(query string, options *QueryOptions) ([]byte, error) {
 	request := &pbuf.QueryRequest{
 		Query:        query,
+		Slices:       options.Slices,
 		ColumnAttrs:  options.Columns,
 		ExcludeAttrs: options.ExcludeAttrs,
 		ExcludeBits:  options.ExcludeBits,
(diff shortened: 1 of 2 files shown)


yuce:
@benbjohnson I think we would need to expose `/slices/max` endpoint to support querying by slices. That endpoint may be removed from the server soon. Is there any emergency about merging this PR? Is it OK to decide about it later?

benbjohnson:
I'm not in a rush but I'll defer to @travisturner.

yuce:
The way of adding options to `client.Query` changed, so current recommended way of passing options is by using `QueryOption` functions, e.g., `client.Query(query, pilosa.ColumnAttrs(true))` instead of `client.Query(query, &pilosa.ClientOptions{Columns: true})`.

yuce:
I think the `QueryOption` for slices would be used something like: `Slices(0, 3)` to use slices 0 and 3.

jaffee:
sorry, I was mistaken, the existing /slices/max endpoint actually already reports the max slice per index.

jaffee:
closing as this is now more up to date in #92

PR #77, opened 2017-11-02
improve support for fields and add some String methods

jaffee:


jaffee:
I added a bit to the integration tests to improve coverage of the field stuff, I don't think we need to be concerned about covering String() functions

client.go
```diff
@@ -286,19 +286,35 @@ func (c *Client) Schema() (*Schema, error) {
...
+			}
+			frameOptions.fields = fields
+			fram, err := index.Frame(frameInfo.Name, frameOptions)
 			if err != nil {
 				return nil, err
 			}
+			for name, _ := range fields {
+				ff := fram.Field(name)
+				if ff.err != nil {
+					return nil, errors.Wrap(err, "fielding frame")
```
yuce: This should wrap `ff.err` instead of `err`. Discovered when adding tests, sending a PR soon.

yuce:
@jaffee Pushed a PR to this PR here: https://github.com/jaffee/go-pilosa/pull/2

PR #80, opened 2017-11-13
Added range field docs to the readme. `RangeField.Sum` accepts `nil` …

yuce:
…for no bitmaps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,51 @@ The recommended way of creating query structs is, using dedicated methods attach
 query := repository.RawQuery("Bitmap(frame='stargazer', row=5)")
 ```
 
+This client supports [Range encoded fields](https://www.pilosa.com/docs/latest/query-language/#range-bsi). Read [Range Encoded Bitmaps](https://www.pilosa.com/blog/range-encoded-bitmaps/) blog post for more information about the BSI implementation of range encoding in Pilosa.
+
+In order to use range encoded fields, a frame should be created with one or more integer fields. Each field should have their minimums and maximums set. Here's how you would do that using this library:
+```go
+index, _ := schema.Index("animals", nil)
+frameOptions := &pilosa.FrameOptions{}
+frameOptions.AddIntField("captivity", 0, 956)
+frame, _ := index.Frame("traits", frameOptions)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -186,6 +186,55 @@ The recommended way of creating query structs is, using dedicated methods attach
 query := repository.RawQuery("Bitmap(frame='stargazer', row=5)")
 ```
 
+This client supports [Range encoded fields](https://www.pilosa.com/docs/latest/query-language/#range-bsi). Read [Range Encoded Bitmaps](https://www.pilosa.com/blog/range-encoded-bitmaps/) blog post for more information about the BSI implementation of range encoding in Pilosa.
+
+In order to use range encoded fields, a frame should be created with one or more integer fields. Each field should have their minimums and maximums set. Here's how you would do that using this library:
+```go
+index, _ := schema.Index("animals", nil)
+frameOptions := &pilosa.FrameOptions{InverseEnabled=true}
```
alanbernstein: `InverseEnabled: true`

README.md
```diff
@@ -186,6 +186,55 @@ The recommended way of creating query structs is, using dedicated methods attach
...
+frameOptions.AddIntField("captivity", 0, 956)
+frame, _ := index.Frame("traits", frameOptions)
+client.SyncSchema(schema)
+``` 
+
+If the frame with the necessary field already exist on the server, you don't need to create the field instance, `client.SyncSchema(schema)` would load that to `schema`. You can then add some data:
```
alanbernstein: exist -> exists

README.md
```diff
@@ -186,6 +186,55 @@ The recommended way of creating query structs is, using dedicated methods attach
...
+	query.Add(captivity.SetIntValue(column, x))
+}
+client.Query(query, nil)
+```
+
+Let's write a range query:
+```go
+// Query for all animals with more than 100 specimens
+response, _ := client.Query(captivity.GT(100), nil)
+fmt.Println("%v", response.Result().Bitmap.Bits)
```
alanbernstein: Printf("%v\n")

yuce:
Thanks!

PR #91, opened 2018-01-03
WIP: Updates for server cluster-resize branch

yuce:
NOTE: not to be merged before `cluster-resize` branch of Pilosa server is merged.

PR #94, opened 2018-01-14
Enterprise Support

benbjohnson:
## Overview

Adds `Key` fields to support translation.


orm.go
```diff
@@ -303,6 +303,11 @@ func (idx *Index) Name() string {
 	return idx.name
 }
 
+// ColumnLabel returns the column label for this index.
+func (idx *Index) ColumnLabel() string {
```
yuce: column/row labels were deprecated in the server and clients and removed in the cluster-resize compatibility PR at: https://github.com/pilosa/go-pilosa/pull/91 (to be merged when the cluster-resize branch is merged.) Do we need support for them in the enterprise server ?

orm.go
```diff
@@ -573,6 +578,11 @@ func (f *Frame) Name() string {
 	return f.name
 }
 
+// RowLabel returns the row label for this frame.
+func (f *Frame) RowLabel() string {
```
yuce: (the same remark with column labels)

orm.go
jaffee: I don't have any issue with dropping them.

benbjohnson:
Closing. Use #95 instead.

PR #96, opened 2018-01-16
add *K ORM methods for enterprise queries

jaffee:


orm.go
```diff
@@ -599,13 +606,27 @@ func (f *Frame) InverseBitmap(columnID uint64) *PQLBaseQuery {
...
 // SetBit creates a SetBit query.
 // SetBit, assigns a value of 1 to a bit in the binary matrix, thus associating the given row in the given frame with the given column.
 func (f *Frame) SetBit(rowID uint64, columnID uint64) *PQLBaseQuery {
 	return NewPQLBaseQuery(fmt.Sprintf("SetBit(%s=%d, frame='%s', %s=%d)",
 		f.options.RowLabel, rowID, f.name, f.index.options.ColumnLabel, columnID), f.index, nil)
 }
 
+// SetBitK creates a SetBit query using string row and column keys. This will
+// only work against a Pilosa Enterprise server.
+func (f *Frame) SetBitK(rowKey string, columnKey string) *PQLBaseQuery {
```
travisturner: This is probably a discussion for another time, but I can imagine a case where a user may only want to use the translator on either Row or Column, but not both.

jaffee: that is a good point... I hope it never happens

PR #98, opened 2018-01-19
WIP: add client.ImportKFrame() method for string keys

travisturner:


client.go
```diff
@@ -413,6 +413,40 @@ func (c *Client) ImportFrame(frame *Frame, bitIterator BitIterator, batchSize ui
 	return nil
 }
 
+// ImportKFrame imports bitKs from the given CSV iterator.
```
jaffee: comment shouldn't mention CSV (also wrong on the original)

yuce:
`ImportFrame`, `ImportValueFrame` (from earlier commits) and `ImportKFrame` are almost the same. I think we have a refactoring opportunity here. The same holds true for `importNode`, `importValueNode` and `importKNode` too.

PR #100, opened 2018-01-19
Fixes augmentHeaders function to prevent data races

### LINKED ISSUE #97: concurrent map access in augmentHeaders
augmentHeaders is called from httpRequest which is often called (in Query for example) with `protobufHeaders` which is a global variable. I've observed a concurrent map access due to this.

We should (a) write a test demonstrating the issue which fails under `go test -race` and (b) fix it... perhaps by augmenting them a single time at startup, or making a copy each time.

yuce:
Adds a data race test. Note that `orm.go` is missing some tests in the master, which will be added in a separate PR.

Fixes #97

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ test:
 	go test
 
 test-all:
-	go test -tags=integration
+	go test -race -tags=integration
 
 release:
 	printf "package pilosa\nconst Version = \"$(VERSION)\"" > version.go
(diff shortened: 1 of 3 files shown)


jaffee:
instead of calling augmentHeaders for every http request, can we set up the headers initially when the client is created, and then use them for each request? In the case that a request actually needs extra headers, we would need to make a copy, but I think that's less common.

yuce:
The `cluster-resize` support branch have some changes in how clients are created, so I didn't want to change any logic there for this PR in order to avoid merge conflicts. Would it be acceptable to revisit this one that branch is merged?

jaffee:
yeah that's totally fine

PR #109, opened 2018-01-24
WIP: Adds QueryResult interface

yuce:


travisturner:
I'd like to know if anyone else has any opinion on this, but it seems strange to me to have methods return valid (default) values when we know, based on `Type`, that those methods are invalid. In other words, I feel like we just shifted the default value into an interface method. I'm wondering if it wouldn't be more useful for a developer calling `TopNResult.Bitmap()` to receive an error?

yuce:
Having default results enable one to write generic functions without checking the type. Let's say a developer wants to write a function that collects attributes:
```
func CollectAttrs(response *QueryResponse) map[string]interface{} {
	res := map[string]interface{}
	for _, result := range response.Results() {
		for k, v := range result.Bitmap().Attributes {
			res[k] = v
		}
	}
}
```
He uses that in the following code:
```
queryString := // comes from somewhere
resp, _ := client.Query(index.RawQuery(queryString))
attrs := CollectAttrs(resp)
```
It should work fine as long as `queryString` contains queries which result in bitmap results. But will panic or fail depending on the result of `result.Bitmap()`. It's easy to protect `CollectAttrs` with a `Type` check, but I think we should not return `nil` as much as we can these kinds of errors. These panics happened to me even when writing tests a few times.

We could return an error for non-BitmapResult results, but then the function signature becomes `Bitmap() (*BitmapResult, error)` which doesn't let us use something like `result.Bitmap().Attributes`.

I think having sane defaults is inline with Go's philosophy, like `var a string` doesn't result in with `a == nil`.

jaffee:
Have `result.Bitmap()` return a `BitmapResult` rather than a `*BitmapResult`. BitmapResult's methods are defined on it rather than a pointer to it anyway, and all of its data is in slices and maps, so it won't be copied.

yuce:
Updated the interface so `BitmapResult` is returned from `BitmapResult()`. Also `BitmapResult.Keys` and `BitmapResult.Bits` always contain an array, even if protobuf unmarshaller sets them to `nil`.

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
 }
 
 func newBitmapResultFromInternal(bitmap *pbuf.Bitmap) (*BitmapResult, error) {
 	attrs, err := convertInternalAttrsToMap(bitmap.Attrs)
 	if err != nil {
 		return nil, err
 	}
+	bits := bitmap.Bits
+	if bits == nil {
+		bits = []uint64{}
```
jaffee: why can't we just leave it nil? Iteration over a nil slice works, len() of a nil slice works, anything accessing a member directly would need to check length first anyway.

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
+	}
+	keys := bitmap.Keys
+	if keys == nil {
+		keys = []string{}
```
jaffee: same question as above.

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
 		Attributes: attrs,
-		Bits:       bitmap.Bits,
-		Keys:       bitmap.Keys,
+		Bits:       bits,
+		Keys:       keys,
 	}
 	return result, nil
 }
 
+func defaultBitmapResult() BitmapResult {
```
jaffee: BitmapResult{} should be the default IMO. Accessing a nil map like `a, ok := map["blah"]` does not panic. Iterating over a nil map does not panic.

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
+func (BitmapResult) Type() uint32 {
+	return QueryResultTypeBitmap
+}
+
+func (b BitmapResult) Bitmap() BitmapResult {
+	return b
+}
+
+func (BitmapResult) CountItems() []*CountResultItem {
+	return []*CountResultItem{}
```
jaffee: just return nil

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
+func (TopNResult) Changed() bool {
+	return false
 }
 
 // BitmapResult represents a result from Bitmap, Union, Intersect, Difference and Range PQL calls.
 type BitmapResult struct {
-	Attributes map[string]interface{}
-	Bits       []uint64
-	Keys       []string
+	Attributes map[string]interface{} `json:"attrs"`
```
jaffee: add omitempty to these

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
+func (NilResult) Type() uint32 {
+	return QueryResultTypeNil
+}
+
+func (NilResult) Bitmap() BitmapResult {
+	return defaultBitmapResult()
+}
+
+func (NilResult) CountItems() []*CountResultItem {
+	return []*CountResultItem{}
```
jaffee: nil

response.go
```diff
@@ -51,10 +52,10 @@ const (
 
 // QueryResponse represents the response from a Pilosa query.
 type QueryResponse struct {
-	ResultList   []*QueryResult `json:"results,omitempty"`
-	ColumnList   []*ColumnItem  `json:"columns,omitempty"`
-	ErrorMessage string         `json:"error-message,omitempty"`
-	Success      bool           `json:"success,omitempty"`
+	ResultList   []QueryResult `json:"results,omitempty"`
+	ColumnList   []*ColumnItem `json:"columns,omitempty"`
```
jaffee: what's the benefit to this being a slice of pointers?

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
+func countItemsFromInternal(items []*pbuf.Pair) TopNResult {
 	result := make([]*CountResultItem, 0, len(items))
 	for _, v := range items {
 		result = append(result, &CountResultItem{ID: v.ID, Key: v.Key, Count: v.Count})
 	}
-	return result
+	return TopNResult(result)
+}
+
+type TopNResult []*CountResultItem
```
jaffee: what's the benefit to this being a slice of pointers?

response.go
```diff
@@ -179,34 +160,228 @@ func (c *CountResultItem) String() string {
...
+
+func (TopNResult) Type() uint32 {
```
jaffee: I'd make all these trivial functions one line and stack them together.
```
func (TopNResult) Type() uint32 { return QueryResultTypePairs }
func (TopNResult) Bitmap() BitmapResult { return BitmapResult{} }
etc.
```

response.go
yuce: if you `json.Marshal` them they become `null` instead of `[]`.

jaffee: I think omitempty will take care of that

yuce: `{}` is not the same as `{bits:[], attrs:[]}`.

jaffee: you're right, hmmm, I was hoping it would treat nil as empty and an empty list as an empty list, but no such luck. 

I'd be in favor of writing a custom MarshalJSON for Bitmap that sets the  defaults before Marshaling - that way we're only setting them in one place, and only if JSON is actually being used. Otherwise we won't incur the extra allocations. We can also handle the sticky business of ensuring that we always have "bits" or "keys", but not both (except in the odd edge case where both have data if we decide to support that).

yuce:
Updated

response.go
```diff
@@ -179,19 +160,45 @@ func (c *CountResultItem) String() string {
...
+	result := make([]CountResultItem, 0, len(items))
 	for _, v := range items {
-		result = append(result, &CountResultItem{ID: v.ID, Key: v.Key, Count: v.Count})
+		result = append(result, CountResultItem{ID: v.ID, Key: v.Key, Count: v.Count})
 	}
-	return result
+	return TopNResult(result)
+}
+
+type TopNResult []CountResultItem
```
jaffee: I think this code would be a lot more readable if we formatted it as follows:
```go
type TopNResult []CountResultItem

func (TopNResult) Type() uint32                    { return QueryResultTypePairs }
func (TopNResult) Bitmap() BitmapResult            { return defaultBitmapResult() }
func (t TopNResult) CountItems() []CountResultItem { return t }
func (TopNResult) Count() int64                    { return 0 }
func (TopNResult) Sum() int64                      { return 0 }
func (TopNResult) Changed() bool                   { return false }
```

yuce:
updated

response.go
```diff
@@ -207,6 +197,93 @@ func newBitmapResultFromInternal(bitmap *pbuf.Bitmap) (*BitmapResult, error) {
...
+}
+
+func (SumCountResult) Type() uint32                  { return QueryResultTypeSumCount }
+func (SumCountResult) Bitmap() BitmapResult          { return BitmapResult{} }
+func (SumCountResult) CountItems() []CountResultItem { return nil }
+func (c SumCountResult) Count() int64                { return c.count }
+func (c SumCountResult) Sum() int64                  { return c.sum }
+func (SumCountResult) Changed() bool                 { return false }
+
+func (c SumCountResult) MarshalJSON() ([]byte, error) {
```
jaffee: do we need these extra MarshalJSON funcs on this and BoolResult?

yuce:
Updated

response.go
```diff
@@ -207,6 +197,82 @@ func newBitmapResultFromInternal(bitmap *pbuf.Bitmap) (*BitmapResult, error) {
...
+		Attributes: b.Attributes,
+		Bits:       bits,
+		Keys:       keys,
+	})
+}
+
+type SumCountResult struct {
+	sum   int64
+	count int64
+}
```
jaffee: Is it possible to change this to
```go
    Sum   int64 `json:"sum"`
    Count int64 `json:"count"`
```

rather than having the separate MarshalJSON method? Or is there somethi

yuce: That's spot on. Updated.

yuce:
Since this PR will break existing code, I've labeled it WIP to prevent it being merged before we update code in dependent projects/communicate with our users.

response.go
yuce: I just remembered. `QueryResult` interface has both `Sum()` and `Count()` functions, so it's not possible to have `Sum`, `Count` fields in `SumCountResult`. That requires adding `MarshalJSON` function to that struct.

jaffee: Oh I see - I would prefer if we just changed the names slightly rather than having the separate MarshalJSON. SumVal and CountVal maybe? then just leave the json tags the same.

yuce: ok

yuce:
Closing in favor of https://github.com/pilosa/go-pilosa/pull/118 (includes this PR)

PR #111, opened 2018-01-26
Adds version checking to master

yuce:


client.go
```diff
@@ -981,6 +1023,17 @@ func TLSConfig(config *tls.Config) ClientOption {
 	}
 }
 
+func SkipVersionCheck(skip bool) ClientOption {
```
jaffee: We should either call this `CheckVersion(check bool)` or `SkipVersionCheck()` (no argument). I think I prefer the latter as it makes it clear that the default behavior is to check the version, but if you pass `SkipVersionCheck()`, the version check will be skipped.

client_it_test.go
```diff
@@ -1502,7 +1502,15 @@ func TestUserAgent(t *testing.T) {
 }
 
 func TestClientRace(t *testing.T) {
-	client := getClient()
+	// Using a custom client with SkipVersionCheck to test race conditions in version checking
+	uri, err := NewURIFromAddress(getPilosaBindAddress())
+	if err != nil {
+		panic(err)
```
jaffee: why panic instead of `t.Fatal(err)` (or similar)?

client.go
```diff
@@ -784,6 +793,38 @@ func (c *Client) statusToNodeSlicesForIndex(status *Status, indexName string) ma
 	return result
 }
 
+func (c *Client) fetchServerVersion() (string, error) {
```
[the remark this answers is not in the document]
yuce: How about calling it `ServerVersion` which returns the cached `version` ?

client.go
```diff
@@ -784,6 +793,38 @@ func (c *Client) statusToNodeSlicesForIndex(status *Status, indexName string) ma
...
+	}
+	versionInfo := versionInfo{}
+	err = json.Unmarshal(data, &versionInfo)
+	if err != nil {
+		return "", errors.Wrap(err, "unmarshaling /version data")
+	}
+	return versionInfo.Version, nil
+}
+
+func checkServerVersion(version string) {
```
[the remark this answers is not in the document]
yuce: ok

client.go
```diff
@@ -755,6 +757,13 @@ func anyError(resp *http.Response, err error) error {
 
 // doRequest creates and performs an http request.
 func (c *Client) doRequest(host *URI, method, path string, headers map[string]string, reader io.Reader) (*http.Response, error) {
+	if !c.versionChecked {
+		// check the server version on the first request
+		c.versionChecked = true
+		// don't care about fetching the version, it's not vital
+		ver, _ := c.fetchServerVersion()
```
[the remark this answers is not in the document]
yuce: The version is fetched the first time a client request is made. Even if the version check fails (any problems fetching it or the version is incompatible) the client request still executes. So even if there is an error fetching the version due to the server is not available or misconfigured the client request will still fail and the user will know about it. I agree that there is a difference between version `""` and a request error. But the first case is handled by `checkVersion` and the second case causes the client request to fail, and the user receives an error.

client.go
yuce: The reason I've called it `SkipVersionCheck` instead of `CheckVersion` is, it sets `SkipVersionCheck` client option directly (which can currently be set by the user too) which in turn is directly assigned to `versionChecked` field in `Client{}` without any logic operators/conditions. `SkipVersionCheck()` without a param looks a bit strange to me. Ideally we could use `SkipVersionCheck` itself, but  that would complicate other parts of the code.

client_it_test.go
yuce: That condition should never fail unless the server address configuration for tests (or `NewURIFromAddress` function) is wrong so checking that error condition is equivalent to no-op in my mind. It also has the benefit of stopping the test run immediately if the server address config is wrong (which would make other tests fail anyway). Admittedly I wasn't consistent with using `t.Fatal`/panics/not checking errors` for `NewURIFromAddress` in the integration tests.

client.go
jaffee: I think if we're going to buy into the whole "functional options" thing then `SkipVersionCheck()` with no arguments is perfectly idiomatic. Ideally, these option funcs would be the only way of setting options and everything else would be unexported. I know that right now we're in the situation of supporting both ways of setting options, but I think we need to deprecate the old way.

client_it_test.go
jaffee: I see - I'm not worried about it now, but `github.com/pilosa/pilosa/test` now has a function for setting up a pilosa cluster in-process, so at some point we should migrate the client integration tests to that rather than having `getPilosaBindAddress()` at all.

client.go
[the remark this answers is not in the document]
jaffee: I'm fine with `ServerVersion` I don't think we should bother with caching it though... if a client is asking for the version, they may be checking that an upgrade of Pilosa has worked, so we'd have to figure out some kind of invalidation strategy. Seems like a pain unless there's a significant performance problem.

client.go
yuce: I agree about deprecating the old way (although it has some benefits over functional options when the options are programmatically generated). I'll change it to `SkipVersionCheck()`, but one potential problem when that flag should be set conditionally depending on the value a variable. Currently the user can just do: `client, _ := pilosa.NewClient(addr, SkipVersionCheck(someVar))`.

client_it_test.go
yuce: That's cool. But wouldn't that make the client tests depend on the Pilosa server?

jaffee: yes it will, is that a problem? having just the tests depend on pilosa won't pull it into the client when used as a library.

client.go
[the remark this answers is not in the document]
jaffee: OK, I agree that it's handled either way, I think it's easier to reason about the code if we check the error though, and it's more robust if someone decides to change how `checkVersion` works and doesn't update this piece of code as well.

yuce:
Updated

yuce:
Closing in favor of https://github.com/pilosa/go-pilosa/pull/118

PR #121, opened 2018-02-28
Added deprecation notices; releasing v0.8.1

### LINKED ISSUE #119: deprecate NewClient* functions and ClientOptions struct
Now that we have the `NewClient` function and we're using functional options, we should deprecate the other functions and options struct to simplify the API. I think at first we can just put some deprecation warnings in there and make sure that all of our tooling is using the new interface, then we can unexport those things so that they aren't exposed anymore.

yuce:
Fixes #119

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ Go client for Pilosa high performance distributed bitmap index.
 * **v0.8.1** (2017-02-28)
     * Compatible with Pilosa 0.8.x and 0.9.x.
     * Checks the server version for Pilosa server compatibility. Passing `SkipVersionCheck()` to `NewClient` disables that.
-    * Supports string row and column IDs. This feature requires Pilosa Enterprise.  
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -12,7 +13,10 @@ Go client for Pilosa high performance distributed bitmap index.
 
 ## Change Log
 
-* **Next**
+* **v0.8.1** (2017-02-28)
+    * Compatible with Pilosa 0.8.x and 0.9.x.
+    * Checks the server version for Pilosa server compatibility. Passing `SkipVersionCheck()` to `NewClient` disables that.
+    * Supports string row and column IDs. This feature requires Pilosa Enterprise.
```
jaffee: We should remove this since we're changing our enterprise strategy. We should probably just say "this feature is not yet supported in Pilosa (so don't use it yet!)"

yuce: I've removed that line altogether to avoid confusion.

yuce:
Great! Merging.

PR #124, opened 2018-04-03
Remove legacy code; remove untestable code; improve coverage

yuce:


yuce:
I thought hard about converting those flags to noop and deprecating them instead of removing them in order to not trivially break dependent code, but decided against in order to not keep superfluous code. But since you think the same about code breakage, I will put the flags back.

yuce:
I've put those flags back, so dependent code shouldn't break trivially.

jaffee:
i was ok with removing them, i just want to avoid adding flags like that
back in the future

client.go
```diff
@@ -1243,6 +966,22 @@ func ExcludeBits(enable bool) QueryOption {
...
+	return func(options *ClientOptions) error {
+		return nil
+	}
+}
+
+// LegacyMode enables legacy mode
+// *DEPRECATED*
+func LegacyMode(enable bool) ClientOption {
+	return func(options *ClientOptions) error {
+		return nil
```
jaffee: Let's add 

```
log.Println("The LegacyMode client option is deprecated and will be removed - it has no effect and should be removed from your code")
```

jaffee:
I'll update the downstream code, and let's make an issue to fully remove the legacy flags after they've been deprecated for a while. I added a comment about adding a log line in LegacyMode to notify users that it's going away.

PR #137, opened 2018-05-18
brings back the old ImportValueFrame logic

jaffee:
Instead of trying to use the same code for importing bits and values, this
brings back the old value importing code. This fixes a bug where the import
manager logic was using the value instead of the column when trying to calculate
the slice for value imports

### DIFF
--- a/client.go
+++ b/client.go
@@ -42,6 +42,7 @@ import (
 	"log"
 	"net"
 	"net/http"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


travisturner:
Is there any value in going the route I suggested here: https://github.com/pilosa/go-pilosa/pull/129#discussion_r186471576

I'll keep reviewing this, but it seems like getting rid of the `index` logic altogether would be better.

jaffee:
I agree - I think more refactoring is needed, but wanted to get a fix out. I have tested this with `pdk gen` and got correct results (and didn't see the wild snapshotting behavior in Pilosa).

client.go
```diff
@@ -367,16 +368,51 @@ func (c *Client) ImportFrame(frame *Frame, iterator RecordIterator, options ...I
...
-	}
-	newOptions := []ImportOption{importBitsFunction(ibf)}
-	for _, opt := range options {
-		newOptions = append(newOptions, opt)
+type ValueIterator interface {
+	NextValue() (FieldValue, error)
+}
+
+// ImportValueFrame imports field values from the given iterator.
+func (c *Client) ImportValueFrame(frame *Frame, field string, valueIterator ValueIterator, batchSize uint) error {
```
travisturner: would there be any value in including the functional options here for batchSize? `options ...ImportOption`

jaffee: maybe - it will still break the PDK unfortunately, so I'll need to update that at the same time.

PR #138, opened 2018-05-19
Refactored Record interface; Removed *Field funcs; Added Slice func.

yuce:


yuce:
There was some discussion about having the ability to support different slice widths in the server. I did some tests with different slice widths and if I am not wrong, I think @jaffee used that as well (?). That option exists to support that. The constant is the default value for `options.sliceWidth`.

PR #140, opened 2018-05-24
change some instances of bit to record to avoid confusion

travisturner:
This PR doesn't change any logic or functionality. It just contains some cases of `bit` changed to `record` that I made while debugging `statusChan` logic in the pdk.


yuce:
There are a few tests failing:
```
./client_it_test.go:762:177: undefined: importBitsFunction
./client_it_test.go:780:163: undefined: importBitsFunction
./client_it_test.go:1605:38: undefined: importBitsFunction
```

yuce:
`client_it_test.go` is compiled only when `tags=integration` flag is given, e.g., with `make test-all`.

PR #142, opened 2018-05-24
prevent posting 0 records in import_manager.go

travisturner:


yuce:
@travisturner Looks good, but the coverage dropped a bit. If your time is tight, I can add a test later.

travisturner:
Thanks @yuce. I will add a test for the additional code path.

travisturner:
@yuce I added a test to bring coverage back up to 100%.
I also discovered some unexpected behavior in `BitGenerator.NextRecord()` where `maxRowID` and `maxColumnID` were treated differently. Previously, `maxRowID=2, maxColumnID=8` returned
```
0: 0,1,2,3,4,5,6,7,8
1: 0,1,2,3,4,5,6,7,8
```
I renamed the arguments to `numRows` and `numColumns`, and now it returns:
```
0: 0,1,2,3,4,5,6,7
1: 0,1,2,3,4,5,6,7
```
That part seems ok, but for some reason the test are sometimes failing on TravisCI by running out of memory (they always pass locally).
I'll keep looking into why that's happening, but I wanted to let you know here what I changed.

yuce:
@travisturner Thanks, these changes look great. I'll also check the out-of-memory problem.

PR #144, opened 2018-05-29
remove all references to inverse and range enabled

jaffee:


Makefile
```diff
@@ -18,10 +18,10 @@ test:
 	go test
 
 test-all:
-	go test -tags=integration -v
+	go test -tags=integration
```
yuce: `-v` flag is useful especially in the CI to see which tests ran.

Makefile
```diff
@@ -18,10 +18,10 @@ test:
...
 
 test-all-race:
-	go test -race -tags=integration -v
+	go test -race -tags=integration
```
yuce: ditto.

client_it_test.go
```diff
@@ -63,7 +63,7 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		panic(err)
 	}
-	testFrame, err = index.Frame("test-frame", &FrameOptions{RangeEnabled: true})
+	testFrame, err = index.Frame("test-frame", &FrameOptions{})
```
yuce: We can leave out passing the frame options in this case.

client_it_test.go
```diff
@@ -1133,8 +1088,7 @@ func TestRangeFrame(t *testing.T) {
 
 func TestCreateIntField(t *testing.T) {
 	client := getClient()
-	options := &FrameOptions{RangeEnabled: true}
-	frame, _ := index.Frame("rangeframe-addfield", options)
+	frame, _ := index.Frame("rangeframe-addfield", &FrameOptions{})
```
yuce: We can leave out passing the frame options in this case.

Makefile
jaffee: in the main pilosa Makefile, we do a TESTFLAGS variable to enable this if necessary. having it on by default is way too much output

jaffee:
updated

yuce:
@jaffee I've pushed a PR to this PR that adds the missing test + `TESTFLAGS`: https://github.com/jaffee/go-pilosa/pull/3

PR #152, opened 2018-06-18
Added OptFieldSet func

yuce:
Cache type and cache size can only be specified through the OptFieldSet function, since they are only valid for set fields.

### DIFF
--- a/orm.go
+++ b/orm.go
@@ -435,9 +435,9 @@ func (fo *FieldOptions) addOptions(options ...interface{}) error {
 // FieldOption is used to pass an option to index.Field function.
 type FieldOption func(options *FieldOptions) error
 
-// OptFieldInt adds a set field.
+// OptFieldSet adds a set field.
 // Specify CacheTypeDefault for the default cache type.
-// Specify 0 for the default cache size.
+// Specify CacheSizeDefault for the default cache size.
 func OptFieldSet(cacheType CacheType, cacheSize int) FieldOption {
 	return func(options *FieldOptions) error {
 		if cacheSize < 0 {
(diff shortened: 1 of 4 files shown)


orm.go
```diff
@@ -438,10 +435,17 @@ func (fo *FieldOptions) addOptions(options ...interface{}) error {
 // FieldOption is used to pass an option to index.Field function.
 type FieldOption func(options *FieldOptions) error
 
-// OptFieldCacheSize sets the cache size.
-func OptFieldCacheSize(size uint) FieldOption {
+// OptFieldInt adds a set field.
```
travisturner: This comment is wrong.

orm.go
```diff
@@ -438,10 +435,17 @@ func (fo *FieldOptions) addOptions(options ...interface{}) error {
...
+// Specify CacheTypeDefault for the default cache type.
+// Specify 0 for the default cache size.
```
travisturner: It seems weird to use the default cache size when the value provided is 0. At that point, why not provide a `CacheSizeDefault` const for consistency with the `CacheTypeDefault ` option? I think if someone sets `cacheSize` to 0 it should effectively act like `CacheTypeNone`.

yuce: Added `CacheSizeDefault` const.

orm.go
yuce: Fixed.

yuce:
@jaffee Right, if cache size is <= 0, i it is not sent to the server side, so the cache size is set to the default on the server side. Line 392 in `orm.go`:
```go
		if fo.cacheSize > 0 {
			mopt["cacheSize"] = fo.cacheSize
		}
```

yuce:
great!

PR #157, opened 2018-07-06
Simple imports refactoring; updated for latest Pilosa develop

yuce:


yuce:
ping!

client.go
```diff
@@ -329,20 +327,18 @@ func (c *Client) importColumns(indexName string, fieldName string, shard uint64,
...
 	for _, node := range nodes {
 		uri := &URI{
 			scheme: node.Scheme,
 			host:   node.Host,
 			port:   node.Port,
 		}
-		eg.Go(func() error {
-			return c.importNode(uri, columnsToImportRequest(indexName, fieldName, shard, records))
-		})
+		err = c.importNode(uri, columnsToImportRequest(indexName, fieldName, shard, records))
```
jaffee: why don't we want to do this concurrently any more?

import_manager.go
```diff
@@ -18,11 +18,17 @@ func newRecordImportManager(client *Client) *recordImportManager {
 	}
 }
 
+type importWorkerChannels struct {
+	recordChan <-chan Record
```
jaffee: why don't we call these fields "records", "errs" and "status" since they'll be referred to like `chans.records` which is a bit cleaner than `chans.recordChan`

client.go
yuce: We run the imports concurrently. The code is at about line 45 of `import_manager.go`. We didn't have the removed code above for `importValues`.

import_manager.go
yuce: That makes sense. Updating.

yuce:
Updated

client.go
jaffee: but this is importing the same data to each of the replica nodes in series. Why wouldn't we want to do that concurrently?

jaffee: I see what you're saying about importValues, but we should probably add it there rather than removing it here - this is doing something serially which is trivially parallel.

yuce: There's no need to send the same data to all nodes which has the shard data. The loop there is supposed to try importing again if there is a problem importing to a node. I'll push a fix soon.

jaffee: are you sure? It doesn't look like the import endpoint automatically replicates the data across shards. the internal client fetches all nodes for a shard and does the import to each of them.

yuce: I see... I assumed that would be the case. Reverting that part of the code then.

yuce: Updated also improved coverage.

PR #159, opened 2018-07-09
expose field options, fix integration tests

jaffee:
when listing fields from the index, one needs to be able to programmtically
determine what type each field is in order to make queries against it.


yuce:
I'll check it again in a few hours but Would it make more sense to have
individual functions instead of exposing field options?

jaffee:
@yuce changed to use methods to expose the options, please see latest.

I considered exposing the options as methods directly on the field type, but ultimately
decided keeping them under a call to `Options()` made more sense. There were some naming conflicts (like Max and Min), and ultimately I decided the API would be less confusing to expose the
options this way.

PR #160, opened 2018-07-16
Fix up index & field key support

benbjohnson:
I updated the client library to fix up some of the index key support so it matched the field support.

### DIFF
--- a/orm.go
+++ b/orm.go
@@ -59,7 +59,7 @@ func NewSchema() *Schema {
 }
 
 // Index returns an index with a name.
-func (s *Schema) Index(name string) (*Index, error) {
+func (s *Schema) Index(name string, options ...interface{}) (*Index, error) {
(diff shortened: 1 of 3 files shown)


yuce:
I've just returned. Reviewing the PR in a bit.

orm.go
```diff
@@ -59,14 +59,21 @@ func NewSchema() *Schema {
 }
 
 // Index returns an index with a name.
-func (s *Schema) Index(name string) (*Index, error) {
+func (s *Schema) Index(name string, options ...interface{}) (*Index, error) {
```
yuce: Would it make sense to make this function to make `options` to take `...IndexOption` and remove the support for `*IndexOptions` ? I will modify `index.Field` to take `...FieldOption`.

yuce:
@benbjohnson Is it OK to merge this PR?

Issue #141, opened 2018-05-24
proposal: remove import strategy option


jaffee:
I think having import strategy be a separate option creates unnecessary API surface - the usual arguments apply here - more to document, more to test, more to support, more potential sources of confusion for users, etc. This, in particular is sticky because it's not necessarily clear how the strategy option interacts with other import options Which other options apply to which strategies? Do some apply to all strategies, or a subset. If you pass an option that doesn't apply to a strategy, will it be silently ignored, or will there be an error?

The batch import strategy is basically what we've been doing and allows the user to tune a pretty straightforward latency/throughput tradeoff. What's missing is having timeouts to import smaller batches in the case that data isn't coming in at a high rate. 

I think we can add this functionality in conjunction with the batch import strategy without having them be mutually exclusive. Whenever a new batch starts, we can start a timer which will be the upper bound on how long to wait before importing that batch. The expected case would be that the data reaches the batch size and is imported well before that timer fires, but in the case that incoming data is slow or sporadic, the timer ensures that data is indexed in a timely fashion and not buffered indefinitely.

travisturner:
My comment here is relevant: https://github.com/pilosa/pdk/issues/77#issuecomment-391846997

travisturner:
I created the following diagram representing my understanding of the `go-pilosa` import logic. The yellow notes indicate two areas where I think there are potential inefficiencies.
![go-pilosa import](https://user-images.githubusercontent.com/410544/40631382-c87d9252-62a1-11e8-9baf-7bd62fdb7923.png)

The diagram below is not a complete proposal, but it indicates some things that I think may improve import efficiency, and hopefully this can spur discussion. The main problem I have with this proposal is that it requires a map lookup (`map[slice]chan Record`) based on `Record.Slice()` for every record. The current solution uses a slice index, which is more efficient. What I do like about this proposal is that `threadCount` refers to the number of workers posting `[]Record` batches to Pilosa. In this case, if records are added in column order, they will still bottleneck through the `sliceChannel`, but there will be multiple workers posting batches concurrently.
![go-pilosa import proposed](https://user-images.githubusercontent.com/410544/40631385-dae3f21a-62a1-11e8-8451-a8059d2c73f8.png)

yuce:
Some thoughts about the proposal:
[...]

PR #161, opened 2018-07-17
[TRIVIAL] Merge develop into master

yuce:


jaffee:
we can't cut 1.0 until we settle the API, and there are several outstanding issues such as https://github.com/pilosa/go-pilosa/issues/141 and https://github.com/pilosa/pilosa/issues/1368

We should at least implement the part of 1368 that simplifies initial setup, and given our discussions around imports, I think we want to remove as many options as possible before releasing that so that we have maximum maneuvering space.

yuce:
@jaffee Is it OK to merge this one?

yuce:
@travisturner @jaffee Can we merge this?

PR #162, opened 2018-07-17
Added missing tests; schema.Index and index.Field refactoring

yuce:
Passing `*IndexOptions` to `schema.Index` and passing `*FieldOptions` to `index.Field` was deprecated before. Removed support for those. Added a few missing tests.


yuce:
ping!

PR #163, opened 2018-07-18
Removes index/field validation. Updated docs.

yuce:
Removed returning `error` value from `schema.Index` and `index.Field` methods.
```go
index := schema.Index("my-index")
field := index.Field("my-field")
```


yuce:
@travisturner Updated.

PR #165, opened 2018-07-24
Adds the support for using column keys with row ids and vice versa

yuce:


orm.go
```diff
@@ -709,6 +693,31 @@ func createAttributesString(attrs map[string]interface{}) (string, error) {
...
+func formatIDKey(idKey interface{}) (string, error) {
+	switch v := idKey.(type) {
+	case uint64:
+		return strconv.FormatUint(v, 10), nil
+	case string:
+		return fmt.Sprintf(`'%s'`, v), nil
+	case int:
+		return strconv.FormatInt(int64(v), 10), nil
+	default:
+		return "", errors.New("Uknown type for id/key")
```
jaffee: errors should not start with a capital letter, and unknown is typo'd.  Let's also include the errant value, so

`errors.Errorf("id/key is not a string or integer type: %#v", idKey)`

orm.go
```diff
@@ -709,6 +693,31 @@ func createAttributesString(attrs map[string]interface{}) (string, error) {
 	return strings.Join(attrsList, ","), nil
 }
 
+func formatIDKey(idKey interface{}) (string, error) {
+	switch v := idKey.(type) {
+	case uint64:
+		return strconv.FormatUint(v, 10), nil
+	case string:
+		return fmt.Sprintf(`'%s'`, v), nil
+	case int:
```
jaffee: we may as well support all integer types 32bit and above, something like:

`case int,int32,int64,uint,uint32,uint64`

yuce:
@jaffee updated

orm.go
```diff
@@ -709,6 +693,39 @@ func createAttributesString(attrs map[string]interface{}) (string, error) {
...
+}
+
+func formatRowColIDKey(rowIDOrKey, colIDOrKey interface{}) (string, string, error) {
+	rowStr, err := formatIDKey(rowIDOrKey)
+	if err != nil {
+		return "", "", errors.Wrap(err, "formatting row")
+	}
+	colStr, err := formatIDKey(colIDOrKey)
+	if err != nil {
+		return "", "", errors.Wrap(err, "formatting row")
```
jaffee: should be "formatting col"

yuce: fixed

PR #166, opened 2018-08-03
Adds support for importing keys

yuce:
This PR works against Pilosa master
~This PR works against: https://github.com/pilosa/pilosa/pull/1557 There are 3 failing tests regarding exports with keys, but those tests should pass once https://github.com/pilosa/pilosa/issues/1580 is resolved.~


yuce:
@jaffee Is it OK to merge this one as well?

client.go
```diff
@@ -489,10 +500,21 @@ func (c *Client) httpRequest(method string, path string, data []byte, headers ma
...
 		reader := bytes.NewReader(data)
-		// get a host from the cluster
-		host := c.cluster.Host()
-		if host == nil {
-			return nil, nil, ErrEmptyCluster
+		var host *URI
+
+		if useCoordinator {
+			// TODO: we shouldn't need to fetch the coordinator every time
+			node, err := c.fetchCoordinatorNode()
```
jaffee: I feel like we should only be fetching the coordinator node if a request fails because it should have been directed to the coordinator, but wasn't.

client.go
```diff
@@ -642,40 +664,101 @@ func makeRequestData(query string, options *QueryOptions) ([]byte, error) {
...
+		rowIDs = make([]uint64, 0, recordCount)
 	}
+
+	if hasColKeys {
+		columnKeys = make([]string, 0, recordCount)
+	} else {
+		columnIDs = make([]uint64, 0, recordCount)
+	}
+
+	if hasRowKeys && hasColKeys {
```
jaffee: I'm thinking this would be more readable as a single `for` loop with two if/else blocks inside (similar to the above two for making the slices).

client.go
yuce: I agree. I thought about adding a `coordinatorNode` field to the `PilosaClient` struct but then I have to add a mutex to prevent races. This is meant to be something that's safe until we find a better solution (like initializing `PilosaClient` with the coordinator node, etc.)

client.go
yuce: I agree, it would be more readable. But I've unrolled the for loop there to make it faster.

imports.go
```diff
@@ -87,24 +89,51 @@ func (b Column) Less(other Record) bool {
 	return false
 }
 
-func ColumnCSVUnmarshaller() CSVRecordUnmarshaller {
-	return ColumnCSVUnmarshallerWithTimestamp("")
+type CSVFormat uint
```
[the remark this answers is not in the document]
yuce: I agree again. Maybe we should remove csv stuff altogether (maybe move that kind of code to its own conveniences repo).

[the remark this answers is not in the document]
jaffee: the pdk might be a good place - it already has some code for csv handling

client.go
yuce: The non-scientific benchmark I've ran with the following script shows that unrolled version is about 1 seconds faster for a batch of 200 million on my machine but simplified this and its field values counterpart anyway.

imports.go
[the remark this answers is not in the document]
yuce: Moved the `csv` stuff to its package but then I had to move some stuff to the `imports` package, since `PilosaClient` depends on the csv stuff for exports (and vice versa). That change increased the lines to be reviewed quite a bit, but most of them are renames, so hopefully that would be trivial.

client.go
jaffee: ok, I agree we can defer further work on this

imports.go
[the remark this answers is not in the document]
jaffee: I'm a bit confused about the exporter - shouldn't it basically just return an io.Reader so that the user can write the exported data to wherever?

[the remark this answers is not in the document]
yuce: I am not sure whether the user would like to write the exported data verbatim, they may transform it or maybe not write it all. Since we keep the CSV stuff in a subpackage, I think it makes sense to use it to make the exported data more useful for the user.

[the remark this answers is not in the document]
jaffee: One could imagine Pilosa's export functionality being a lot more general than just dumping a CSV. It could support multiple content types - perhaps even in a pluggable way. I think that the responsibility of decoding that data falls outside the go-pilosa package. It also goes against the package structure that we've generally adopted to have the core package depend on subpackages - we've striven to make the reverse true in Pilosa.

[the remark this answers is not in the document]
yuce: OK, ditched the imports package, so the core doesn't depend on it, but that requires refactoring the integration tests. Hold on...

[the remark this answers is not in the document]
jaffee: sorry :/

yuce:
Refactoring the integration tests will take longer than I assumed. Have to
finish it later.

yuce:
@jaffee Removed the imports package, the PR is good to review again.

yuce:
ping!

client.go
```diff
@@ -489,17 +491,40 @@ func (c *Client) httpRequest(method string, path string, data []byte, headers ma
...
+				c.coordinatorURI = coordinatorURI
+				c.coordinatorLock.Unlock()
+			}
+			host = coordinatorURI
+		} else {
+			// get a host from the cluster
+			host = c.cluster.Host()
+			if host == nil {
+				return nil, nil, ErrEmptyCluster
+			}
```
jaffee: I think the logic in this whole if/else block should be wrapped up in cluster.Host and Host should be modified to take a boolean `useCoordinator` argument.

client.go
```diff
@@ -489,17 +491,40 @@ func (c *Client) httpRequest(method string, path string, data []byte, headers ma
...
+			c.coordinatorLock.RLock()
+			coordinatorURI := c.coordinatorURI
+			c.coordinatorLock.RUnlock()
+			if coordinatorURI == nil {
+				node, err := c.fetchCoordinatorNode()
+				if err != nil {
+					return nil, nil, errors.Wrap(err, "fetching coordinator node")
+				}
+				coordinatorURI = URIFromAddress(fmt.Sprintf("%s://%s:%d", node.Scheme, node.Host, node.Port))
+				c.coordinatorLock.Lock()
```
jaffee: I *think* this lock should be acquired before calling `fetchCoordinatorNode` to avoid multiple goroutines trying to fetch the coordinator simultaneously. This also needs to do a final check of c.coordinatorURI after acquriring the write lock in case it was set by another goroutine between releasing the read lock and acquiring the write lock.

client.go
yuce: I've thought about moving that logic to the cluster (probably `cluster.Host`) before but that creates a chicken-egg problem: we can't get the coordinator node without creating the client first; and creating the client requires a cluster. That if/else code is messy though, I'll move it to a separate function.

client.go
yuce: I've locked the mutex after fetching the coordinator node to prevent other goroutines block reading the coordinatorURI, but I am not sure my logic in that code is correct. I'm adding a test to see whether anything blocks.

client.go
```diff
@@ -521,6 +527,33 @@ func (c *Client) httpRequest(method string, path string, data []byte, headers ma
...
 }
 
+func (c *Client) host(useCoordinator bool) (*URI, error) {
+	var host *URI
+	if useCoordinator {
+		c.coordinatorLock.RLock()
+		coordinatorURI := c.coordinatorURI
+		c.coordinatorLock.RUnlock()
+		if coordinatorURI == nil {
+			node, err := c.fetchCoordinatorNode()
```
jaffee: I believe that we need to obtain the write lock before calling `fetchCoordinatorNode`, my reasoning is as follows:

multiple routines may call Client.host simultaneously. They might simultaneously obtain a read lock and see that the coordinator is nil, they would then all try to fetch the coordinator node simultaneously which would be wasteful since only one needs to do it. Then they will all try to obtain the write lock to set the coordinator node to what they found, and each one will succeed in succession.

What we should do is obtain the write lock, then *re-check that c.coordinatorURI is still nil*, if it is still nil, then proceed with fetchCoordinator node. If it was set in between releasing the RLock and obtaining the Lock, then we can release the lock and use it.

yuce:
@jaffee I've updated the PR with your feedback and a test. I'll need to make the test more robust, other  than that it should be OK to review.

yuce:
@jaffee  I've reverted the coordinator lock logic, so it locks after fetching the coordinator node. Without that, `client.host` would sometimes return `nil` which causes a panic. I didn't get that problem once I've reverted the logic.

jaffee:
It was returning nil because there was a bug - if you unrevert it, I'll comment on where the change needs to be.

yuce:
unreverted

client.go
```diff
@@ -521,6 +527,36 @@ func (c *Client) httpRequest(method string, path string, data []byte, headers ma
...
 
+func (c *Client) host(useCoordinator bool) (*URI, error) {
+	var host *URI
+	if useCoordinator {
+		c.coordinatorLock.RLock()
+		coordinatorURI := c.coordinatorURI
+		c.coordinatorLock.RUnlock()
+		if coordinatorURI == nil {
+			c.coordinatorLock.Lock()
+			if c.coordinatorURI == nil {
```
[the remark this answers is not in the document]
yuce: How did I miss that??? Thanks for pointing that out. Updated.

yuce:
I've simplified the coordinator lock logic a bit; removed the `coordinatorURI == nil` check after the lock. Rarely `coordinatorURI != nil` and the latest `else` we've added is not testable (since it occurs rarely).

jaffee:
now it's possible for many goroutines to get piled up on the write lock, and then each will re-fetch the coordinator node in series.

jaffee:
for what it's worth, I'm fine with that line not being covered reliably. As long as we have a test which calls `host` concurrently in several goroutines, gets the right answer, and passes the race detector, I'm satisfied that it's sufficiently covered.

yuce:
Yeah, but leaving that line drops the coverage. I want to try a few things before I give up.

jaffee:
I figured as much ;-) just letting you know my position

yuce:
@jaffee Improved the coverage.

yuce:
Added back to `nil` check too.

PR #168, opened 2018-08-14
Sends PQL-Version to server; logs warnings from server

yuce:


yuce:
Since Go doesn't support anything other than `log.Print` or `log.Fatal`, we won't be able to support error logs (unless we use a 3rd party logger). I think `log.Fatal` is too extreme, but we can return an error.

PR #169, opened 2018-08-21
Direct queries with keys to coordinator

yuce:
This is a PR to https://github.com/pilosa/go-pilosa/tree/import-enhancements

### DIFF
--- a/client.go
+++ b/client.go
@@ -52,7 +52,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-const PqlVersion = "1.0"
+const PQLVersion = "1.0"
 const maxHosts = 10
 
 // Client is the HTTP client for Pilosa server.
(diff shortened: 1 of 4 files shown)


orm.go
```diff
@@ -112,26 +112,44 @@ func (s *Schema) diff(other *Schema) *Schema {
 	return result
 }
 
+type SerializedQuery struct {
```
jaffee: does this need to be exported?

orm.go
```diff
@@ -112,26 +112,44 @@ func (s *Schema) diff(other *Schema) *Schema {
 	return result
 }
 
+type serializedQuery struct {
+	Query   string
+	HasKeys bool
+}
+
+func NewSerializedQuery(query string, hasKeys bool) serializedQuery {
```
jaffee: this function should be unexported as well, I think

client.go
```diff
@@ -50,6 +52,7 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const PqlVersion = "1.0"
```
jaffee: this should be named PQLVersion according to Go naming conventions

orm.go
yuce: Unexported.

client.go
yuce: Renamed

PR #170, opened 2018-08-22
Fix TopN filter variant

yuce:


yuce:
yeah, I wouldn't change the function name otherwise.

PR #172, opened 2018-09-04
[TRIVIAL] Make generates renames the protobuf definition package

yuce:


Makefile
```diff
@@ -12,6 +12,7 @@ fast-cover:
 	go test -cover -tags="integration" $(TESTFLAGS)
 
 generate:
+	sed -i -e 's/package internal;/package gopilosa_pbuf;/g' gopilosa_pbuf/public.proto
```
jaffee: I don't understand why this needs to be in the Makefile. Isn't it sufficient to make the change once to the .proto file and then run `make generate`?

yuce: Yeah, but we copy the proto file from the Pilosa repo, and we may need to update it manually. We may forget to update the proto file too.

jaffee: oh, I see - good idea. Can you add a comment to make that clear?

yuce: Sure, updated.

Makefile
```diff
@@ -12,6 +12,8 @@ fast-cover:
 	go test -cover -tags="integration" $(TESTFLAGS)
 
 generate:
+	# Rename the package name in the proto definition to match the package
```
jaffee: This says what it does, but that was already fairly clear. It still isn't clear why this is necessary. I want it to say something like "this ensures that we don't forget to change the package name if we copy the proto definition from pilosa in order to update it. It is important that the packages have different names because proto can misbehave if pilosa and go-pilosa are imported into the same codebase."

kuba--:
Thanks, would be cool to release it as 0.10.0 (or something)

yuce:
@kuba-- The 0.10.0 release is here: https://github.com/pilosa/go-pilosa/tree/v0.10.0

PR #167, opened 2018-08-08
Parameterize integration tests, test multiple versions of Pilosa

codysoyland:
Move integration test orchestration into Makefile
Simplify .travis.yml
Add multiple Pilosa versions to build matrix

Issue #158, opened 2018-07-09
integration tests should use MustRunCluster and not rely on external Pilosa


jaffee:
github.com/pilosa/pilosa/test exposes a "MustRunCluster" which is useful for creating and destroying temporary clusters to test against. We no longer need to rely on an externally running Pilosa instance for client integration tests.

Furthermore, we should remove the `// +build integration` tags so that all tests run on every test run. The build tags sometimes break automated code analysis and refactoring tools (such as gorename which ignores them by default), so removing them will simplify things.

codysoyland:
I spent some time on this and got it working on one test, but I realized that using MustRunCluster would lock us into only testing one Pilosa version, and it would be impossible to run integration tests coupling different versions of the client and server.

I decided instead to make it easier to run the test suite with a temporary Pilosa server by automating that in the Makefile. I also added a Travis CI matrix configuration to test different versions of Pilosa. PR is here: https://github.com/pilosa/go-pilosa/pull/167

codysoyland:
I determined that a better path for this would be to switch to CircleCI (as we have for Pilosa server), which allows running the integration tests in separate containers and makes it easier to parameterize. This is now merged (PR #179).

PR #179, opened 2018-09-18
CircleCI build

codysoyland:
This makes some changes to the Makefile and the CI process. The Makefile now includes integration test orchestration tooling, and multiple Pilosa versions can now be run against the integration tests in CI. This also adds gometalinter. CircleCI improves speed with parallel tests and caching.

This replaces #167 and serves as an alternative to #158.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,19 @@
+.PHONY: all cover fast-cover generate-proto test test-all test-all-race gometalinter
 
-.PHONY: all cover fast-cover generate-proto test test-all
-
+PILOSA_VERSION ?= master
+PILOSA_VERSION_ID := pilosa-$(PILOSA_VERSION)-$(GOOS)-amd64
+PILOSA_DOWNLOAD_URL ?= https://s3.amazonaws.com/build.pilosa.com/$(PILOSA_VERSION_ID).tar.gz
 VERSION := $(shell git describe --tags 2> /dev/null || echo unknown)
+GOOS ?= linux
+PILOSA_BIND ?= http://:10101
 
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,30 +1,72 @@
...
 all: test
 
-cover:
-	go test ./... -cover -tags="integration fullcoverage" $(TESTFLAGS)
+cover: vendor
+	PILOSA_BIND=$(PILOSA_BIND) go test ./... -cover $(TESTFLAGS) -covermode=count -coverprofile=build/coverage.out
 
 fast-cover:
-	go test ./... -cover -tags="integration" $(TESTFLAGS)
+	PILOSA_BIND=$(PILOSA_BIND) go test ./... -cover -tags="nointegration" $(TESTFLAGS)
```
yuce: I think we should remove `nointegration` tag, so integration tests run for this command.

Makefile
```diff
@@ -1,30 +1,72 @@
...
 VERSION := $(shell git describe --tags 2> /dev/null || echo unknown)
+GOOS ?= linux
+PILOSA_BIND ?= http://:10101
 
 all: test
 
-cover:
-	go test ./... -cover -tags="integration fullcoverage" $(TESTFLAGS)
+cover: vendor
+	PILOSA_BIND=$(PILOSA_BIND) go test ./... -cover $(TESTFLAGS) -covermode=count -coverprofile=build/coverage.out
```
yuce: We need `fullcoverage` tag here, otherwise *heavy* coverage tests don't run.

PR #184, opened 2018-09-26
Roaring imports are disabled by default. Added OptImportRoaring option.

yuce:


yuce:
@codysoyland I've added the `PREVENT_ROARING` environment variable, which suppresses setting the `roaring` flag to `true`. We would run Pilosa 1.1 instances with `PREVENT_ROARING=true` and `master` branch with `PREVENT_ROARING=false`. Would that work ?

jaffee:
that's pretty ugly - I'm -1 on adding an environment variable.

Would it be possible to have the client make a request to /import-roaring on startup to determine whether the endpoint is supported?

yuce:
@jaffee This change doesn't affect the client at all, only two integration tests.

yuce:
To be more clear; we have a new `OptImportRoaring` option which controls whether roaring imports is enabled (disabled by default). Two of the integration tests sets that flag depending on the `PREVENT_ROARING` env var.

jaffee:
I'm not sure I agree with having roaring import be an option at all. I'd say let's see if Pilosa supports it and use it if it's there. Won't make a difference to the client, right? (except for being faster)

yuce:
OK, the client will test whether roaring import endpoint is available before importing.

yuce:
Updated

codysoyland:
I like the recent changes, but it is failing in CI. Would it pass with Pilosa master?

yuce:
Cicle CI build fails because Pilosa 1.1 doesn't have the `trackExistence` index option . Travis CI build runs on Pilosa master and fails due to a panic, but strangely doesn't panic on local. As far as I see, it should panic on local too. Trying to find out why it doesn't.

yuce:
I've checked the code again, and there's nothing wrong with it (I thought since `p.index` is `nil` in `TestMakeRequestData` a panic would be thrown, but the line that would panic is never executed since the marshalling fails) . The error line number is different between runs on travis ci, I think that's because of a problem on travis ci's side (out of memory, etc.).

PR #193, opened 2018-10-26
implements clear imports

yuce:


client.go
```diff
@@ -999,6 +1004,13 @@ type ImportOptions struct {
 	statusChan            chan<- ImportStatusUpdate
 	importRecordsFunction func(field *Field, shard uint64, records []Record, nodes []fragmentNode, options *ImportOptions) error
 	roaring               bool
+	clear                 bool
+}
+
+func DefaultImportOptions() *ImportOptions {
```
travisturner: Does this need to be exported?
I've been trying to avoid allowing a user to provide the `ImportOptions` struct directly. Instead, the end user interface should be functional options. In the core, I had to export the `ImportOptions` struct because that's the struct that the `OptImportOptions*` functional options operate on, but it's not used as an input to any exported methods.

yuce: You're right, it is only used by the tests. Moved it to the integration test.

PR #196, opened 2018-11-01
Address minor documentation bitrot.

seebs:
README.md had references to PilosaException, probably
inherited from the java-pilosa documentation. Also,
CONTRIBUTING.md had the wrong URL for the contribution
guide in the main Pilosa repo (lacking the '/blob/master'
which is needed to look at a specific file).


yuce:
@seebs Can we merge this PR?

jaffee:
@seebs 	our process is typically that the submitter of the PR is the one who merges it (after approval, assuming the submitter has write access). I think you should have write access, but if not, let me know.

PR #207, opened 2018-12-05
initial client-side row key to id translation, RowKeyColID roaring import

yuce:
It's not perfect yet, but it works as far as the tests go.


yuce:
This should be good to review. The coverage is dropped a bit, but I am adding more tests in a separate PR.

client.go
```diff
@@ -303,40 +305,71 @@ func (c *Client) Schema() (*Schema, error) {
...
+	options *ImportOptions,
+	state *importState) error {
 	eg := errgroup.Group{}
 
 	if len(nodes) == 0 {
 		return errors.New("No nodes to import to")
 	}
 
-	if options.roaring {
+	if options.hasRoaring {
```
seebs: This is starting to look large/complicated enough to be a separate function, maybe?

yuce: thanks, updated

yuce:
Yeah, the cache size cannot be less than the batch size in order to protect against the case that all keys in a batch are new. Need to add this to the docs.

yuce:
I think enabling multi-pass translation in order to keep the cache size low is a good idea. It makes the translation a bit more complex and potentially slower (multiple translate calls to the server) but it could help if the batch size is very large (100m or so?) but if the import driver has little memory to spare (for translation cache). I'll think a bit about that.

PR #211, opened 2018-12-15
performance tweaks

seebs:
This is some low-hanging fruit for improving client runtime. Actual wall-clock time is barely affected by this, which makes me suspect that some of that is server-side, but I haven't diagnosed it more clearly, I just wanted to get a bunch of the easy things out there for review.

Before:
```
    real    0m43.058s
    user    1m12.904s
    sys     0m1.072s
```

After:
```
    real    0m40.063s
    user    0m29.443s
    sys     0m0.888s
```

This certainly frees up a lot of CPU time, but the impact on wall-clock time is trivial thus far.

### DIFF
--- a/client.go
+++ b/client.go
@@ -405,7 +405,9 @@ func (c *Client) importColumns(field *Field,
 		return c.importColumnsRoaring(field, shard, records, nodes, options, state)
 	}
 
-	sort.Sort(recordSort(records))
+	if !options.skipSort {
+		sort.Sort(recordSort(records))
+	}
 
 	for _, node := range nodes {
 		uri := node.URI()
@@ -1317,6 +1319,7 @@ type ImportOptions struct {
 	clear              bool
 	rowKeyCacheSize    int
 	columnKeyCacheSize int
+	skipSort           bool
 }
 
 func (opt *ImportOptions) withDefaults() (updated ImportOptions) {
@@ -1380,6 +1383,23 @@ func OptImportClear(clear bool) ImportOption {
 func OptImportRoaring(enable bool) ImportOption {
 	return func(options *ImportOptions) error {
 		options.wantRoaring = enable
+		return nil
+	}
+}
+
+// OptImportSort tells the importer whether or not to sort batches of records, on
+// by default. Sorting imposes some performance cost, especially on data that's
+// already sorted, but dramatically improves performance in pathological
+// cases. It is enabled by default because the pathological cases are awful,
+// and the performance hit is comparatively small, but the performance cost can
+// be significant if you know your data is sorted.
+func OptImportSort(sorting bool) ImportOption {
+	return func(options *ImportOptions) error {
+		// skipSort is expressed negatively because we want to
+		// keep sorting enabled by default, so the zero value should
+		// be that default behavior. The client option expresses it
+		// positively because that's easier for API users.
+		options.skipSort = !sorting
 		return nil
 	}
 }
--- a/import_manager.go
+++ b/import_manager.go
@@ -128,6 +128,9 @@ readRecords:
 	for record := range recordChan {
 		recordCount++
 		shard := record.Shard(shardWidth)
+		if batchForShard[shard] == nil {
+			batchForShard[shard] = make([]Record, 0, batchSize)
+		}
 		batchForShard[shard] = append(batchForShard[shard], record)
 
 		if recordCount >= batchSize {
@@ -139,7 +142,7 @@ readRecords:
 				if err != nil {
 					break readRecords
 				}
-				batchForShard[shard] = nil
+				batchForShard[shard] = batchForShard[shard][:0]
 			}
 			recordCount = 0
 		}
--- a/import_manager.go
+++ b/import_manager.go
@@ -21,7 +21,7 @@ func newRecordImportManager(client *Client) *recordImportManager {
 }
 
 type importWorkerChannels struct {
-	records <-chan Record
+	records <-chan []Record
 	errs    chan<- error
 	status  chan<- ImportStatusUpdate
 }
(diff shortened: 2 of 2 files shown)


seebs:
Update: I dropped the pointer-receiver column implementation, because it was a lot of effort and API complexity for a marginal performance improvement in a case that can be avoided by just using the "don't sort" option added in a later patch. I kept the record-batching, but dropped the exposed setting for it; instead, I just set it to 16, which seems to be pretty good.

On further study, the reason this didn't affect wall clock time is performance issues on the server side, which I'm addressing separately in the main tree. My actual wall clock time changed from ~55 seconds to ~15 seconds for the test case I was doing once I figured that out. :)

yuce:
As far as I've tested, the decrease in total import time was marginal  (as you also mentioned). I've run the following tests using 1000 rows, 10_000 columns data with `RowIDColumnID` format on a single node. I've attached the import script to the ticket. I've repeated the tests a bunch of times, but the numbers below are representative:

Using this PR with sorting disabled (`OptImportSort(false)`):

```
$ time ./importer-s-nosort :10101 /media/ext/yuce/TEMP/ri1000_ci10000_r.csv 100000 false true true
field keys? false
2019/01/09 18:27:23 Took: 18012 milliseconds

real	0m18.502s
user	0m12.340s
sys	0m0.928s
```

Using this PR with sorting enabled:
```
9$ time ./importer-s :10101 /media/ext/yuce/TEMP/ri1000_ci10000_r.csv 100000 false true true
field keys? false
2019/01/09 18:28:14 Took: 18700 milliseconds

real	0m19.216s
user	0m14.258s
sys	0m0.545s
```

Using the PR at: https://github.com/pilosa/go-pilosa/pull/207 with roaring import enabled:
```
$ time ./importer-y :10101 /media/ext/yuce/TEMP/ri1000_ci10000_r.csv 100000 false true true
field keys? false
2019/01/09 18:28:55 Took: 11876 milliseconds

real	0m12.324s
user	0m19.957s
sys	0m0.412s
```

Using the PR at: https://github.com/pilosa/go-pilosa/pull/207 with roaring import disabled:
```
$ time ./importer-y :10101 /media/ext/yuce/TEMP/ri1000_ci10000_r.csv 100000 false true false
field keys? false
2019/01/09 18:29:47 Took: 20082 milliseconds

real	0m20.549s
user	0m18.110s
sys	0m0.651s
```

I've tested with smaller columnwise data and the results were similar. I think we''ll have the biggest import improvement in import time with roaring imports, which  will also remove the need for sorting.

I'll try to test with different data to see whether there's some dramatic gains with this PR.

The importer script (roaring-related lines were uncommented for the roaring test):
```go
package main

import (
	"fmt"
	"log"
	"os"
	"strconv"
	"time"

	pilosa "github.com/pilosa/go-pilosa"
	"github.com/pilosa/go-pilosa/csv"
)

func main() {
	if len(os.Args) != 7 {
		log.Fatalf("Usage: %s Pilosa_addr csv_file batch_size row_keys? col_keys? roaring?", os.Args[0])
	}

	addr := os.Args[1]
	csvPath := os.Args[2]
	batchSize, err := strconv.Atoi(os.Args[3])
	rowKeys := os.Args[4] == "true"
	colKeys := os.Args[5] == "true"
	// wantRoaring := os.Args[6] == "true"

	picl, err := pilosa.NewClient(addr)
	if err != nil {
		log.Fatal(err)
	}

	f, err := os.Open(csvPath)
	if err != nil {
		log.Fatal(err)
	}
	defer f.Close()

	schema, err := picl.Schema()
	if err != nil {
		log.Fatal(err)
	}
	index := schema.Index("i1", pilosa.OptIndexKeys(colKeys))
	field := index.Field("f1", pilosa.OptFieldKeys(rowKeys))
	err = picl.SyncSchema(schema)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Println("field keys?", field.Options().Keys())

	csvFormat := csv.RowIDColumnID
	if rowKeys && !colKeys {
		csvFormat = csv.RowKeyColumnID
	} else if !rowKeys && colKeys {
		csvFormat = csv.RowIDColumnKey
	} else if rowKeys && colKeys {
		csvFormat = csv.RowKeyColumnKey
	}
	iter := csv.NewColumnIterator(csvFormat, f)

	tic := time.Now()
	err = picl.ImportField(field, iter,
		pilosa.OptImportBatchSize(batchSize),
		pilosa.OptImportThreadCount(2),
		pilosa.OptImportSort(false),
		// pilosa.OptImportRoaring(wantRoaring),
	)
	if err != nil {
		log.Fatal(err)
	}
	took := time.Since(tic)
	log.Printf("Took: %d milliseconds\n", uint64(took.Nanoseconds()/1000000))
}
```

seebs:
A bit after I initially submitted this, I did some profiling server-side, cleaned up some code on that side, and discovered that this made a much larger difference in wall clock time. But most of my changes were around the BSI/Value code, not the raw individual bits code.
 
With my test program (which is generating data, not using CSV):

```
before:
real    0m7.302s
user    0m11.630s
sys     0m0.232s

after:
real    0m2.171s
user    0m4.183s
sys     0m0.113s
```

Using your test case, I don't see nearly as much of a performance difference. More significantly, the real time is much larger than user time, which means the program's waiting most of the time:

```
real    1m5.304s
user    0m18.264s
sys     0m0.462s
```

So that implies server-side is most of the delay there. I'm experimenting more to explore this.

seebs:
so it turns out that the case i was testing there was an accidental pathological case in which I did about 10M rows of data on 10 columns.

If I write the data I *meant* to write, I get:

```
before:
real    0m13.149s
user    0m16.578s
sys     0m0.268s

after:
real    0m7.660s
user    0m15.213s
sys     0m0.472s

after, without sort:
real    0m4.648s
user    0m7.095s
sys     0m0.224s
```

So that's some benefit. But that's also with unusually-easy data.

PR #213, opened 2018-12-24
Adds Rows and GroupBy calls

yuce:


yuce:
@jaffee I've added decoding responses for `Rows` and `GroupBy` queries, seems I've missed those before. Added integration tests for those queries.

PR #218, opened 2019-02-20
Added OptClientManualServerAddress option

yuce:
Fixes: https://github.com/pilosa/pilosa/issues/1857
~TODO: integration test~


client.go
```diff
@@ -668,12 +710,12 @@ func (c *Client) httpRequest(method string, path string, data []byte, headers ma
 	if data == nil {
 		data = []byte{}
 	}
+	reader := bytes.NewReader(data)
```
seebs: I am not sure why this moved up here, but I think this could fail when maxHosts is greater than 1, because we pass reader in to c.doRequest, and I think we need to make a new reader for each call, or the later ones will all get empty bodies?

yuce: Thanks, you are right, I've carelessly "optimized" that code. Moved it back now.

yuce:
I've added more tests.

PR #219, opened 2019-02-25
goreportcard fixes

yuce:


client_it_test.go
```diff
@@ -999,8 +1002,22 @@ func TestRangeField(t *testing.T) {
 	if resp.Result().Count() != 2 {
 		t.Fatalf("Count 2 != %d", resp.Result().Count())
 	}
+}
+
+func TerstRangeField2(t *testing.T) {
```
seebs: Typo: Should probably be "Test".

csv/csv.go
```diff
@@ -100,14 +110,17 @@ func NewIterator(reader io.Reader, unmarshaller RecordUnmarshaller) *Iterator {
 	}
 }
 
+// NewColumnIterator creates a new iterator for column data
```
seebs: This, and a couple of nearby comments, lack terminal `.`, but I think the convention is for complete sentences including punctuation.

import_manager.go
```diff
@@ -90,10 +90,9 @@ sendRecords:
 	return importErr
 }
 
-func recordImportWorker(id int, client *Client, field *Field, chans importWorkerChannels, options ImportOptions) {
+func recordImportWorker(id int, client *Client, field *Field, chans importWorkerChannels, options *ImportOptions) {
```
seebs: Not that I have an objection, but why did ImportOptions become a pointer here? Is this just because in general convention is to use pointers for structs that aren't specifically known to be very small?

orm.go
```diff
@@ -1016,16 +1021,27 @@ func formatRowColIDKey(rowIDOrKey, colIDOrKey interface{}) (string, string, erro
...
+	// See: https://www.pilosa.com/docs/latest/data-model/#int
+	FieldTypeInt FieldType = "int"
+	// FieldTypeTime is the time field type.
+	// See: https://www.pilosa.com/docs/latest/data-model/#time
+	FieldTypeTime FieldType = "time"
+	// FieldTypeMutex is the mutex field type.
+	// See: https://www.pilosa.com/docs/latest/data-model/#mutex
+	FieldTypeMutex FieldType = "mutex"
+	// FieldTypeBool is the boolean field type.
+	// See: https://www.pilosa.com/docs/latest/data-model/#boolean
```
seebs: I like the links for the specific field types. Possibly we should also have such a link on the FieldType description?

client_it_test.go
yuce: Thanks, updated.

csv/csv.go
yuce: Thanks, updated.

orm.go
yuce: Thanks, added the link.

import_manager.go
yuce: Yeah, I thought not-copying `ImportOptions` would be fine there since it is not mutated anywhere. I had to move `importRecords` to its global function, since goreportcard was complaining that `recordImportWorker` became too complex.

yuce:
To be honest quite a bit of the doc comments are there to make the linter happy. If there are any that we can improve, we probably should do that.

yuce:
I think `Re-request review` action is new, clicking on it to see what happens...

yuce:
Great, github didn't crash...

PR #220, opened 2019-03-01
index shard width is loaded from the server

yuce:
It works, but need to modify some of the imports tests to improve coverage


yuce:
CI build fails since I guess it doesn't yet use the latest Pilosa master. I'll rerun the CI build in 30.

yuce:
This should be good to review

yuce:
Thanks @seebs . I'll add an option to enable auto-reading the shard width.

yuce:
If the shard width cannot be loaded from the server, the client uses the default shard width on import. That should solve the new client/old server problem.

PR #221, opened 2019-03-20
Added hasIndex and HasField functions

yuce:


yuce:
That's possible using something like:
```go
// check whether f1 exists and is an int field
var f1IsIntField bool
if index.HasField("f1") {
  f1 := index.Field("f1")
  if f1.Opt().Type() == OptFieldTypeInt {
    f1IsIntField = true
  }
}
```

seebs:
Yeah. What I sort of want, but don't see an easy way to do client-side, is check whether a field matches a given set of field options. This is hard to do client-side, because so far as I can tell, the actual default values and combinations and error-checking is done server-side. But I didn't realize that we had the `.Opt()` thing -- I should probably read the code more carefully. So that looks fine to me for now.

PR #223, opened 2019-03-29
Exported query serialization

yuce:
It is useful to peek at what a query is for debugging, etc. This PR exports the query serialization interface.


yuce:
great!

PR #231, opened 2019-04-24
Key translation updates

### LINKED ISSUE #226: make import-roaring the default
We talked about this a bit in https://github.com/pilosa/pdk/issues/117 but the more I think about, the more I think it makes sense for import-roaring to be the default. 

It's enormously faster, and avoids a lot of memory pressure on Pilosa for time fields. Since turning it on is really an implementation detail from the perspective of a user of the client library, I don't see much downside to doing it.

On the other hand, if we leave it off by default, users of go-pilosa will see bad performance by default.

Since we're probably going to move everything in the direction of roaring imports and deprecate regular imports, it would make sense to me to exercise them by default.

yuce:
* Defer creating key translation caches until required.
* Included simplelru in the code and adapted it to make it better work with the client: string keys and uint64 values; removed eviction handler.
* Roaring import is the default. Fixes #226

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,9 +51,9 @@ import (
 	"time"
 
 	"github.com/golang/protobuf/proto"
-	lru "github.com/hashicorp/golang-lru/simplelru"
 	"github.com/opentracing/opentracing-go"
 	pbuf "github.com/pilosa/go-pilosa/gopilosa_pbuf"
+	"github.com/pilosa/go-pilosa/lru"
 	"github.com/pilosa/pilosa/roaring"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
@@ -481,7 +481,7 @@ func (c *Client) importColumnsRoaring(field *Field,
 		// attempt to translate row keys
 		if state.rowKeyIDMap == nil {
 			// create the row key cache if it doesn't exist
-			state.rowKeyIDMap, err = lru.NewLRU(options.rowKeyCacheSize, nil)
+			state.rowKeyIDMap, err = lru.NewLRU(options.rowKeyCacheSize)
 			if err != nil {
 				panic(errors.Wrap(err, "while creating rowKey to ID map"))
 			}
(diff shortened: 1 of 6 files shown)


import_manager.go
```diff
@@ -122,19 +121,7 @@ func recordImportWorker(id int, client *Client, field *Field, chans importWorker
...
-	columnKeyIDMap, err := lru.NewLRU(options.columnKeyCacheSize, nil)
-	if err != nil {
-		panic(errors.Wrap(err, "while creating columnKey to ID map"))
-	}
-	state := &importState{
-		rowKeyIDMap:    rowKeyIDMap,
-		columnKeyIDMap: columnKeyIDMap,
-	}
-
+	state := &importState{}
```
seebs: I'm not sure I fully understand this. It looks like the impact of this is that we don't create the caches until the first time a given worker uses them, but other than that, it should behave the same (reusing them for the lifetime of that worker, etc.)?

client.go
```diff
@@ -397,7 +397,7 @@ func (c *Client) ImportField(field *Field, iterator RecordIterator, options ...I
 			importRecordsFunction(c.importValues)(&importOptions)
 		} else {
 			// Check whether roaring imports is available
-			if importOptions.wantRoaring {
+			if importOptions.wantRoaring != nil && *importOptions.wantRoaring == true {
```
seebs: we could possibly drop this, and the section around 1500, by treating nil as true. so `if importOptions.wantRoaring == nil || *importOptions.wantRoaring == true`, maybe?

import_manager.go
yuce: This allows us to postpone cache creation is postponed until it is required, like rowKey cache is only created for `RowKeyColumnID` or `RowKeyColunKey` type data; it's similar for columnKey cache. Related code is in `importColumnsRoaring` in `client.go`.

client.go
yuce: I thought about moving the default check out of `withDefaults` but that forces us to remember the default, which may be problematic if we check against `importOptions.wantRoaring` elsewhere.

seebs: Hmm. That's true. So in this case, the != nil isn't intended as part of the decision logic, it's a sanity-check to prevent panics if something went wrong elsewhere?

yuce: Sorry, I've merged the PR before I saw your comment. Yeah, I've just added it for piece of mind.

PR #232, opened 2019-04-26
add configurable concurrency for import replay

jaffee:
I saw a 4x speedup importing locally.


client.go
```diff
@@ -1022,45 +1022,66 @@ func (c *Client) logImport(index, path string, shard uint64, isRoaring bool, dat
...
-			resp.Body.Close()
+			return nil
+		})
+	}
+
+	// populate work channel
+	dec := newImportLogDecoder(r)
+	var err error
+	for {
+		l := importLog{}
```
yuce: The same comment about `l` here.

jaffee:
good feedback—renamed that

PR #236, opened 2019-05-01
Initial retry support

### LINKED ISSUE #233: retry requests when temporary  errors occur
Pilosa clients need a way to distinguish between temporary and permanent errors when querying Pilosa. 

If a temporary error occurs, the client should retry the request a few times, sleeping successively longer intervals between retries.

The motivating use case for this is when a Pilosa cluster goes into state STARTING briefly because it thinks a node is down due to a hiccup in memberlist. This is often a transient condition and it would be convenient if it didn't cause things like imports to fail.

yuce:
Fixes #233 
TODO: Need to do more tests...

### DIFF
--- a/client.go
+++ b/client.go
@@ -598,8 +598,9 @@ func (c *Client) hasRoaringImportSupport(field *Field) bool {
 	}
 	// Check whether the roaring import endpoint exists
 	path := makeRoaringImportPath(field, 0, url.Values{})
-	resp, _, err := c.httpRequest("GET", path, nil, nil, false)
-	if err != nil {
+	// err may contain an HTTP error, but we don't use it.
+	resp, _, _ := c.httpRequest("GET", path, nil, nil, false)
+	if resp == nil {
 		return false
 	}
 	if resp.StatusCode == http.StatusMethodNotAllowed || resp.StatusCode == http.StatusOK {
@@ -968,9 +969,9 @@ func (c *Client) doRequest(host *URI, method, path string, headers map[string]st
 		}
 	}
 	if err != nil {
-		return nil, errors.Wrap(err, "max retrials exceeded")
+		return nil, errors.Wrap(err, "max retries exceeded")
 	}
-	return nil, errors.New("max retrials exceeded")
+	return nil, errors.New("max retries exceeded")
 }
 
 // statusToNodeShardsForIndex finds the hosts which contains shards for the given index
--- a/client.go
+++ b/client.go
@@ -82,6 +82,10 @@ type Client struct {
 	manualFragmentNode *fragmentNode
 	manualServerURI    *URI
 	tracer             opentracing.Tracer
+	// Number of retries if an HTTP request fails
+	retrials          int
+	minRetrySleepTime time.Duration
+	maxRetrySleepTime time.Duration
 
 	importLogEncoder encoder
 	logLock          sync.Mutex
@@ -143,6 +147,9 @@ func newClientWithOptions(options *ClientOptions) *Client {
 	} else {
 		c.tracer = options.tracer
 	}
+	c.retrials = 0
+	c.minRetrySleepTime = 1 * time.Second
+	c.maxRetrySleepTime = 2 * time.Minute
 	c.importManager = newRecordImportManager(c)
 	return c
 
@@ -591,7 +598,10 @@ func (c *Client) hasRoaringImportSupport(field *Field) bool {
 	}
 	// Check whether the roaring import endpoint exists
 	path := makeRoaringImportPath(field, 0, url.Values{})
-	resp, _, _ := c.httpRequest("GET", path, nil, nil, false)
+	resp, _, err := c.httpRequest("GET", path, nil, nil, false)
+	if err != nil {
+		return false
+	}
 	if resp.StatusCode == http.StatusMethodNotAllowed || resp.StatusCode == http.StatusOK {
 		// Roaring import endpoint exists
 		return true
@@ -680,7 +690,7 @@ func (c *Client) fetchCoordinatorNode() (fragmentNode, error) {
 }
 
 func (c *Client) importData(uri *URI, path string, data []byte) error {
-	resp, err := c.doRequest(uri, "POST", path, defaultProtobufHeaders(), bytes.NewReader(data))
+	resp, err := c.doRequest(uri, "POST", path, defaultProtobufHeaders(), data)
 	if err = anyError(resp, err); err != nil {
 		return errors.Wrap(err, "doing import")
 	}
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -82,6 +82,10 @@ type Client struct {
 	manualFragmentNode *fragmentNode
 	manualServerURI    *URI
 	tracer             opentracing.Tracer
+	// Number of retries if an HTTP request fails
+	retrials          int
```
seebs: would probably use "retries" rather than "retrials"

client.go
```diff
@@ -925,12 +935,40 @@ func anyError(resp *http.Response, err error) error {
...
+	var err error
+	var req *http.Request
+	retrial := 1 + c.retrials
+	sleepTime := c.minRetrySleepTime
+	for retrial > 0 {
+		req, err = makeRequest(host, method, path, headers, data)
+		if err != nil {
+			return nil, errors.Wrap(err, "building request")
+		}
+		retrial -= 1
```
seebs: i think one of the linters will ask for "--" here.

client.go
```diff
@@ -925,12 +935,40 @@ func anyError(resp *http.Response, err error) error {
 }
 
 // doRequest creates and performs an http request.
-func (c *Client) doRequest(host *URI, method, path string, headers map[string]string, reader io.Reader) (*http.Response, error) {
-	req, err := makeRequest(host, method, path, headers, reader)
+func (c *Client) doRequest(host *URI, method, path string, headers map[string]string, data []byte) (*http.Response, error) {
```
seebs: why the change from io.Reader to []byte? oh, so it can be reused. my concern is that this would prevent streaming. but then, any io.Reader could prevent us from retrying if it isn't seekable or something.

client.go
yuce: ok, updated.

client.go
yuce: updated.

client.go
yuce: Yeah, this seems like the easiest approach to support retries. I don't understand why it does prevent streaming though. Can't the caller of this function read some data from the reader to a byte array?

yuce:
I've updated the PR with default retries to be 2. Also `400` status is retried too. It should be ready.

client.go
```diff
@@ -925,12 +935,42 @@ func anyError(resp *http.Response, err error) error {
...
+		}
+		tries--
+		resp, err = c.client.Do(req)
+		if err == nil {
+			if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+				return resp, nil
+			}
+			// Pilosa nodes sometimes return 400, we retry in that case.
+			// No need to retry in other 4xx cases.
+			if resp.StatusCode > 400 && resp.StatusCode < 500 {
```
jaffee: I think we probably need to be more careful in Pilosa about what statuses we're returning, and which errors are retryable, and which are not. There are many cases with 400 errors where retrying doesn't make sense (unmarshal error at `pilosa/http/handler.go:1100`), and I think in some places we're returning 5XX errors, where it would make sense to retry (see `api.Import` at `pilosa/http/handler.go:1104`).

So... let's decide on one (or a set of) status codes that signify a temporary error condition where it's reasonable for the client to perform a retry on its own, and then let's make sure that Pilosa returns those status codes in the appropriate cases and does *not* return them when a retry doesn't make sense.

I think that the logic as it stands right now, won't actually retry if an import-roaring request fails due to the cluster being in state STARTING which is the motivating use case for this work.

client.go
```diff
@@ -925,12 +935,42 @@ func anyError(resp *http.Response, err error) error {
...
+		}
+		time.Sleep(sleepTime)
+		sleepTime *= 2
+		if sleepTime > c.maxRetrySleepTime {
+			sleepTime = c.maxRetrySleepTime
+		}
+	}
 	if err != nil {
-		return nil, errors.Wrap(err, "building request")
+		return nil, errors.Wrap(err, "max retrials exceeded")
```
jaffee: "retries" rather than "retrials" IMO

client.go
seebs: With the io.Reader form, the caller can pass an arbitrarily-large stream to a single request, without needing to allocate storage. This caps a request at transmitting a block of data that fits comfortably in memory, and requires users to know how much data they want to send, and handle making sequential requests for sequential blocks -- and also to know how to divide the incoming data into blocks, and handle any header-like things that have to go in front of them, etcetera.

client.go
yuce: Actually it seems like it works quite nice when one of the nodes is in `STARTING` state. The instructions for reproducing https://github.com/pilosa/pilosa/issues/1585 in the #dev channel leaves the coordinator in `STARTING` state, and the client keeps retrying the import request. When the other node is restarted, the coordinator becomes `NORMAL` and the import progresses.

I agree with updating Pilosa to return better error codes BUT it's not a very big problem currently. Even if retrying doesn't make sense in some cases, it's also harmless.

We retry 500 errors.

client.go
yuce: updated

client.go
yuce: The biggest requests are import requests, and for roaring imports we create bitmaps which should be in memory so at least for roaring imports probably we can't support streaming (without allocating). Also sending the request  body in chunks wouldn't make a difference for the server, so reading a buffer from an `io.Reader` and posting it would work, no need to deal with headers, etc.

client.go
jaffee: ah, I missed that we retry on 500... that makes more sense

client.go
jaffee: My $0.02 is that I'm not worried about this right now. Column-major roaring imports aren't going to stream very well, and streaming isn't supported in Pilosa since (e.g.) unmarshalBinary takes a []byte.

I would like to tackle this sort of holistically at some point, but I think the way things are now this makes sense.

seebs: fair enough. it's a conceptual objection but i'm not sure i have an actual use case for it.

yuce:
@jaffee @seebs Is it OK to merge this PR?

seebs:
I don't object, but I'm insufficiently confident that I understand the surrounding context to say yes.

client.go
```diff
@@ -925,12 +936,42 @@ func anyError(resp *http.Response, err error) error {
...
+		}
+		time.Sleep(sleepTime)
+		sleepTime *= 2
+		if sleepTime > c.maxRetrySleepTime {
+			sleepTime = c.maxRetrySleepTime
+		}
+	}
 	if err != nil {
-		return nil, errors.Wrap(err, "building request")
+		return nil, errors.Wrap(err, "max retries exceeded")
```
jaffee: if the `err` is nil down here, then we'll fail with "max retries exceeded" and no information about what problem was actually occurring. It seems like this could happen in the expected case of a 400 or 5XX when the cluster is in state STARTING. Similarly, the log message above gives the status code rather than an actual error message which would be more informative.

Probably the easiest thing would be to see if the err from the request is nil, and if it is, but the status is 400 or 5XX to set err to a new error which includes the message from the response. Then use a single logger line, and a single return line down here since the error should be guaranteed to be non-nil.

yuce: updated.

PR #239, opened 2019-06-07
Fixes #238

### LINKED ISSUE #238: Do not roaring import if trackExistence is true
Currently `trackExistence` is not handled on the client-side on roaring import. So roaring import should be disabled when the index has `trackExistence=true` until we support it on the client-side or server-side.

yuce:
- Adds Field.HasRoaringImport method
- Disables roaring imports when trackExistence os true

### DIFF
--- a/client.go
+++ b/client.go
@@ -405,7 +405,7 @@ func (c *Client) ImportField(field *Field, iterator RecordIterator, options ...I
 		} else {
 			// Check whether roaring imports is available
 			if importOptions.wantRoaring != nil && *importOptions.wantRoaring == true {
-				importOptions.hasRoaring = c.hasRoaringImportSupport(field)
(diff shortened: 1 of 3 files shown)


jaffee:
Instead of making this check on the client side, I wonder if we should...
(a) disable this server side, or
(b) allow it, and also expose the ability to roaring import into the existence field. Basically, if you're using roaring imports, and existence tracking, you're responsible for updating the existence field, which in most cases can probably be done with very little overhead by importing full RLE containers in your known column range. 

I'm leaning toward (b), though this probably requires a bit of work in Pilosa.

yuce:
Both approaches are possible, but this PR is more like a very quick way to avoid confusion for go-pilosa users.

PR #248, opened 2019-07-08
Add MinRow and MaxRow for fields

### LINKED ISSUE #247: Add MinRow and MaxRow for fields
Pilosa now supports MinRow and MaxRow calls in https://github.com/pilosa/pilosa/pull/1983. We should also add support for this in the Go client.

shaqque:
This PR attempts to add MinRow and MaxRow calls to fields in an attempt to fix #247. Unfortunately, the tests to verify this works don't pass.

### DIFF
--- a/client_it_test.go
+++ b/client_it_test.go
@@ -327,12 +327,12 @@ func TestMinMaxRow(t *testing.T) {
 
 	response, err := client.Query(field.MinRow())
 	if err != nil {
-		t.Fatalf("error excecuting min: %v", err)
+		t.Fatalf("error executing min: %v", err)
 	}
 	min := response.Result().Value()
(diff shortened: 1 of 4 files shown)


shaqque:
The tests now pass with the changes in #249.

client_it_test.go
```diff
@@ -307,6 +307,44 @@ func TestTopNReturns(t *testing.T) {
 	}
 }
 
+func TestMinMaxRow(t *testing.T) {
+	client := getClient()
+	field := index.Field("test-field")
```
yuce: Can you change the field name to something like `test-minmaxrow-field`? It seems we've used `test-field` elsewhere and that makes this test fail.

shaqque: done!

PR #251, opened 2019-08-15
add ability to add to lru cache without eviction, fix ingest bug

jaffee:
go-pilosa keeps an LRU cache of the row key to id mapping so that it
doesn’t have to look it all up from Pilosa every time. It first checks
the cache to see which keys it needs to look up, then looks up from
Pilosa and adds to the cache. Unfortunately, it was possible for the
newly added keys to evict ones which were previously in the
cache. When it went to use them, it couldn’t find them and errored
out, which eventually caused ingest to deadlock (when using PDK).


yuce:
hmm, the LRU is supposed to be created with the size of the `batchSize` so it should have enough capacity without evicting. I'll check this deeper.

jaffee:
Yes, but if the cache has size 1000, and a batch comes in with 500 new keys and 500 existing keys, the 500 new keys will be looked up and added. While they are being added, 500 existing keys will be removed, and if there is any overlap between the removed keys and the keys which were existing when we first checked, then some of the keys which were originally existing will not be present when the cache is used later in the function.

PR #253, opened 2019-08-27
Batch ingest

jaffee:
Experimental batch ingestion interface. The godocs hopefully make things clear. See https://github.com/pilosa/pdk/pull/130 for some example usage.

Performance-wise, this seems to be far more efficient (at least 5x) than the current import interface, and can probably do even better with some profiling and tuning.


cmd/picsv/main.go
```diff
@@ -0,0 +1,323 @@
...
+
+	// IDField denotes which field in the source should be used for Pilosa record IDs.
+	IDField string `json:"id-field"`
+
+	// IDType denotes whether the ID field should be parsed as a string or uint64.
+	IDType string `json:"id-type"`
+}
+
+type Field struct {
+	Type      string           `json:"type"`
```
tgruben: Using the term "type" causes some problems in some java config file parsers/tools because it is a reserved word.  Just a minor annoyance to consider

jaffee:
@tgruben can you approve this one so I can merge it before the PDK stuff?

gpexp/importbatch.go
```diff
@@ -0,0 +1,743 @@
...
+
+func (b *Batch) doImport() error {
+	eg := errgroup.Group{}
+
+	frags, err := b.makeFragments()
+	if err != nil {
+		return errors.Wrap(err, "making fragments")
+	}
+	for shard, fieldMap := range frags {
+		for field, viewMap := range fieldMap {
```
dmibor: had quite a bit of time to play around with this way of ingesting on our data - maybe this info would be useful for you guys:

this way of ingesting a fixed size batch would heavily bottleneck when number of shards grows and data that is coming to the batch start to be distributed over more and more shards and therefore fragments. Eventually what happened with out data is that doesnt mapper how small batch is configured - we get 1-2 most recent shards in a batch that have most of the data in them, about 70% just in them, they are big. And then we get 100s and 100s of tiny shards. And importing these 100s of tiny shards to Pilosa would heavily bottleneck on unioning bitmaps on the Pilosa cluster, since data read from disk would inevitably happen.

That is what hard limit on batch size does when IDs repeat themselves, how bad it would be depends on the pattern of how IDs repeat themselves in the data, for us it was extremely bad.

What proved to work for us eventually - is to go more granular, put hard limit not to a batch, but to every single fragment. I.e. import fragment to Pilosa as soon as it reaches certain size and flush it from memory. Plus limit number tiny fragment bitmaps in memory by time - to avoid memory blowing up from thousands of tiny fragment bitmaps.

This tremendously reduces and in fact puts an upper bound on number of unions Pilosa cluster has to do during import irrespective of the shard size. It made things a lot faster

dmibor: maybe you can consider implementing streaming ingestion when rows are streamed into go-pilosa API and are flushed to Pilosa cluster periodically when above conditions are met

dmibor: or maybe I can submit my code of this logic to go-pilosa if you guys are interested in something like that

jaffee: @dmibor thanks for the input!

This experiment is definitely skewed toward making the easier cases really fast rather than handling the more difficult scattered/semi-random ID problem. 

We have some Pilosa work going on which is making it more efficient to send small updates to Pilosa, and will have to see how that affects things.

I would be interested to understand how you decided to limit the number of small fragments in memory - you said "by time", but I'm not sure I understand if that's like oldest, or LRU or... 

FWIW, this batch is fixed by number of records, not number of bits in a given shard, so no matter how scattered the IDs are, there is a hard upper limit on memory usage.

dmibor: Yeah, I do see that this way go-pilosa memory footprint would be stable.

"by time" meaning flush small fragments every N minutes, i.e. for every fragment if can be flushed if 1) it's too big and bit count is more than some config, say 1M bits 2) if it is sitting in memory more than N configured minutes - this timeout makes sure tiny fragments sent to Pilosa not too often and are buffered in memory of go-pilosa instead as opposed to fixed batch way of sending when tiny fragments are send as often as batch fills up, multiple times per second at least.

On the other hard, that also puts memory footprint under control, fragment bitmaps can not become too big or there can not be too many of fragments to exhaust memory.

PR #255, opened 2019-10-16
prevent nil pointer exception when pilosa cluster does not exist

travisturner:


client.go
```diff
@@ -1127,7 +1127,11 @@ func (c *Client) doRequest(host *URI, method, path string, headers map[string]st
 		if tries == 0 {
 			break
 		}
-		c.logger.Printf("request failed with: %v status: %d, retrying after %d more time(s) after %v ", err, resp.StatusCode, tries, sleepTime)
+		var statusCode int
+		if resp != nil {
+			statusCode = resp.StatusCode
+		}
+		c.logger.Printf("request failed with: %v status: %d, retrying after %d more time(s) after %v ", err, statusCode, tries, sleepTime)
```
alanbernstein: Since we're here anyway:

```suggestion
		c.logger.Printf("request failed with: %v status: %d, retrying %d more time(s) after %v ", err, statusCode, tries, sleepTime)
```

jaffee: quote the '%v' before status as well... that is really confusing — it makes it look the error message ends with the word "status".

(I wrote this code)

travisturner: I made those changes

PR #256, opened 2019-10-16
adds a DecodeFieldOptions function to reverse the FieldOptions stringer

travisturner:


travisturner:
There's a problem with this when the json integer values are large. The integer option `min` and `max` are an example. Since they are decoded as `float64`, but are actually `int64`, we lose precision when the values are, for example, `math.MinInt64` and `math.MaxInt64`.

travisturner:
This does not have to be in go-pilosa. Moving it to the app.

PR #258, opened 2019-10-25
Don't use batch size for channel buffers

seebs:
Channel buffers are there to let workers do a bit more work
while things are running, but batch size is for the amount of stuff
we want to accumulate before dropping a whole batch of computed
results off with the server. There's no reason for the channel buffer
to be that size, and having it be that size (which is typically at
least 65536) means that each import worker is typically sitting on
a channel that has about a million interface-flavored objects containing
pointers to records in it, and suddenly we have memory pressure.

(Fixes, I hope, #4.)

### DIFF
--- a/import_manager.go
+++ b/import_manager.go
@@ -128,6 +128,9 @@ func recordImportWorker(id int, client *Client, field *Field, chans importWorker
 
 readRecords:
 	for recordBatch := range recordChan {
+		// It's fine to overrun our allowed batch size slightly, and
+		// we don't want to generate separate batches for part of a
+		// 16-item batch.
 		for _, record := range recordBatch {
 			recordCount++
 			shard := record.Shard(shardWidth)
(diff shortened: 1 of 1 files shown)


seebs:
FWIW: a workload that was getting OOM-killed somewhere past 30GB of memory before this is staying under 1GB of memory after this. This is the mysterious memory leak I was pretty sure I'd detected in imagine, but hadn't fixed yet.

tgruben:
I say this in uber's go styleguide, https://github.com/uber-go/guide/blob/master/style.md#channel-size-is-one-or-none.  It went against my intuition, but i kinda makes sense to me.

seebs:
There's some utility to small-positive channel sizes, although mostly as an implementation of semaphores, but they can be useful for buffering to provide some insulation against bursty things. But we definitely don't need 65k+ sized channels.

